PR #54, opened 2022-10-28
feat: make combobox filter prop configurable

abhishekv24:


packages/apsara-ui/src/Combobox/Combobox.tsx
```diff
@@ -25,6 +25,7 @@ const Combobox = ({
     showSearch = true,
     showArrow = true,
     placeholder,
+    optionFilterProp = "value",
```
StewartJingga, line 28: this method is actually on works for `undefined`. if user passes `null` or `""` it won't be changed to `"value"`.

packages/apsara-ui/src/Combobox/Combobox.tsx
```diff
@@ -82,7 +83,7 @@ const Combobox = ({
             onDeselect={onValueDeselect}
             options={options}
             filterOption={true}
-            optionFilterProp="label"
+            optionFilterProp={optionFilterProp}
```
StewartJingga, line 86:
```suggestion
            optionFilterProp={optionFilterProp || 'value'}
```

we can do the default thing here

StewartJingga (changes requested):

StewartJingga (approved):

PR #55, opened 2022-11-10
feat: migrate modal, loader, states, segments, search, diff timeline, content layout, header from antd

abhishekv24:


abhishekv24:
Table component changes are also in this PR with story named Tanstack table.
Once the changes are complete I would replace the current table with that.

packages/apsara-ui/src/Segments/Segments.tsx
```diff
@@ -1,16 +1,19 @@
...
     span?: number;
     children?: React.ReactNode;
 }
 
-const Segments = ({ span = 10, children }: SegmentsProps) => <Col span={span}>{children}</Col>;
+const calculateWidth = (span: number) => {
+    return (span / 24) * 100 + '%'
+}
+
+const Segments = ({ span = 10, children }: SegmentsProps) => <div style={{ maxWidth: calculateWidth(span) }}> {children}</ div>;
```
StewartJingga, line 16: do we really need span here? or is it just internal use?

I feel like we can skip `span`, CMIIW it was there because we use `Col` as the internal component for `Segment`?

packages/apsara-ui/src/Skeleton/Skeleton.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import { ListSkeleton, EditorSkeleton, DetailsSkeleton } from "./Skeleton";
+
+export default {
+    title: "Feedback/Skeleton",
```
StewartJingga, line 5: should we rename this to `Loading` instead of `Skeleton`? Seems more intuitive, or is `Skeleton` a common term?

packages/apsara-ui/src/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,50 @@
...
+        </StyledSkeleton>
+    )
+}
+
+interface EditorSkeletonProps {
+    className?: string;
+    width?: string;
+    lastLineWidth?: string
+}
+export function EditorSkeleton({ className = "", width = "100%", lastLineWidth = "100%" }: EditorSkeletonProps) {
```
StewartJingga, line 24: wdyt about `Form*` instead of `Editor*`? Will that improve it?

StewartJingga (changes requested):

packages/apsara-ui/src/Segments/Segments.tsx
abhishekv24: I retained span here just to preserve the backward compatibility, 
with `Col` being used previously span was being used there as well just to set the width.

packages/apsara-ui/src/Skeleton/Skeleton.stories.tsx
abhishekv24: I suppose `Loading/Loader` alone is rather confusing, CMIIW it gives sense of that circular loading animation we have.
`Skeleton` is rather a common term for this one and I think it is being used in many other common UI libraries.
@StewartJingga maybe we can rename it to `SkeletonLoading` or `SkeletonLoader`.
@ravisuhag @AkarshSatija What do you guys suggest for naming this one?

packages/apsara-ui/src/Skeleton/Skeleton.tsx
abhishekv24: yeah we can definitely rename it, I just kept the previously used names.

packages/apsara-ui/src/DiffTimeline/DiffTimeline.stories.tsx
```diff
@@ -0,0 +1,73 @@
+import React from "react";
+
+import DiffTimeline from "./DiffTimeline";
+
+export default {
+    title: "General/DiffTimeline",
+    component: DiffTimeline,
+};
+
+const dataArr = [
```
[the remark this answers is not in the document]
abhishekv24: Yeah sure, I can add that.

packages/apsara-ui/src/Header/Header.styles.tsx
```diff
@@ -1,15 +1,68 @@
...
+export const AvatarFallback = StyledFallback;
+
+
+export const Wrapper = styled.div`
     display: flex;
     align-items: center;
     padding: 0 40px;
     background-color: ${({ theme }) => theme?.header?.menuBg};
     border-bottom: solid 1px ${({ theme }) => theme?.colors?.black[3]};
+    line-height: 64px
```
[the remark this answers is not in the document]
abhishekv24: ok I will try that.

packages/apsara-ui/src/Modal/Modal.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import Button from "../Button";
+import { Overlay, Content, Title, HeadingWrapper, CloseDiv } from "./Modal.styles";
+import { CloseOutlined } from "@ant-design/icons";
```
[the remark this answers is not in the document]
abhishekv24: oh I missed this one.
Yes radix has a similar icon we can use here.

packages/apsara-ui/src/Skeleton/Skeleton.stories.tsx
StewartJingga: just did a quick research, and Skeleton seems to be pretty common. We can keep it.

packages/apsara-ui/src/Header/Header.styles.tsx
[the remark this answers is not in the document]
abhishekv24: replaced line-height with height.

packages/apsara-ui/src/Modal/Modal.tsx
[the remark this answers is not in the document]
abhishekv24: replaced the antd icon with radix icon

packages/apsara-ui/src/DiffTimeline/DiffTimeline.stories.tsx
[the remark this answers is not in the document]
abhishekv24: @StewartJingga can you please help me out with this one?
I got data of other cases from daggers history page but I am not able to find any for deleted case.

StewartJingga (approved):

PR #60, opened 2022-12-01
fix: apsara 0.8.0 issue fixes

abhishekv24:


ravisuhag (approved):

packages/apsara-ui/package.json
```diff
@@ -52,10 +52,18 @@
         "@radix-ui/react-avatar": "^1.0.0",
         "@radix-ui/react-dropdown-menu": "^1.0.0",
         "@tanstack/react-table": "8.5.13",
+        "@rc-component/mini-decimal": "1.0.1",
         "react-query": "3.39.2",
         "react-virtual": "^2.10.4",
         "rc-pagination": "^3.1.17",
         "antd": "^4.16.13",
```
StewartJingga, line 59: So if we remove this, we would need to import lots of packages ?

Does that mean Apsara depend on libraries that are `antd`'s dependencies?

packages/apsara-ui/src/Button/Button.types.tsx
```diff
@@ -1,8 +1,61 @@
...
+    | "right"
+    | "bottom"
+    | "topLeft"
+    | "topRight"
+    | "bottomLeft"
+    | "bottomRight"
+    | "leftTop"
+    | "leftBottom"
+    | "rightTop"
+    | "rightBottom";
```
StewartJingga, line 57: Just a question, are these all antd props?

packages/apsara-ui/package.json
abhishekv24: removed this dependency as we were using only one type from there.

packages/apsara-ui/src/Button/Button.types.tsx
abhishekv24: Yes, ButtonProps and TooltipPlacement types were being imported from antd here.

ravisuhag:
@StewartJingga @abhishekv24 i am merging this PR. if they can other small issues we can cover them in another one.

PR #104, opened 2024-05-21
feat: add datatable calendar filter

rsbh:


packages/raystack/calendar/range-picker.tsx
```diff
@@ -16,31 +16,38 @@ interface RangePickerProps {
...
 
   const handleSelect: SelectRangeEventHandler = (range, selectedDay) => {
-    if (range) {
-      setValue(range);
-    }
-    if (range?.to === selectedDay) {
-      onOpenChange(false);
-      onSelect(range);
+    let from = value?.from || range?.from;
+    from = dayjs(from).isAfter(dayjs(selectedDay)) ? undefined : from;
```
anujk14, line 44: Couldn't quite understand the logic from 43-50. I'm assuming it has to do with "auto-populating" to and from depending on what the user has clicked. Can you explain the expected behaviour we are targeting once?

Thinking out loud if there is scope to make this more readable/understandable

packages/raystack/table/FilteredChip.tsx
```diff
@@ -29,26 +30,43 @@ interface FilterValuesProps {
...
-      : filterValueTypeMap.text;
+  const [value, setValue] = useState<FilterValue>({
+    date: new Date(),
+    dateRange: {
+      to: new Date(),
+      from: new Date(),
+    },
+  });
+
+  const valueType: filterValueType = filterOperation?.component
```
anujk14, line 51: Can we use an if..else here? Nested ternary operations can be difficult to debug and maintain later on when we look at it after a while.

packages/raystack/table/FilteredChip.tsx
```diff
@@ -29,26 +30,43 @@ interface FilterValuesProps {
...
 
   useEffect(() => {
-    if (value && onValueChange) {
-      onValueChange({ value });
+    if (onValueChange) {
+      onValueChange(value);
     }
   }, [value]);
+
   return valueType === filterValueTypeMap.select ? (
```
anujk14, line 65: For readability of this entire section, can we use a switch case here, and get the different components we return via functions?
```
switch(valueType) {
  case filterValueTypeMap.select {
     return getSelectFilter(arg1, arg2, arg3, ....)
  }
  case filterValueTypeMap.datePicker {
    return getDatePicker(arg1.....)
  }
}
```

Actually, we could even use the factory pattern here, depending on the types of inputs we expect to receive.

packages/raystack/table/filterFns.tsx
```diff
@@ -123,4 +130,45 @@ export const operationsOptions: Record<columnTypes, Array<FilterOperation>> = {
...
+      fn: (row, columnId, filterValue: FilterValue) => {
+        return dayjs(row.getValue(columnId)).isAfter(dayjs(filterValue.date));
+      },
+    },
+    {
+      label: "is between",
+      value: "is between",
+      fn: (row, columnId, filterValue: FilterValue) => {
+        return (
+          dayjs(row.getValue(columnId)).isAfter(
```
anujk14, line 163:
```suggestion
          dayjs(row.getValue(columnId)).isSameOrAfter(
```

packages/raystack/table/filterFns.tsx
```diff
@@ -123,4 +130,45 @@ export const operationsOptions: Record<columnTypes, Array<FilterOperation>> = {
...
+            dayjs(filterValue.dateRange?.from)
+          ) &&
+          dayjs(row.getValue(columnId)).isBefore(
```
anujk14, line 166:
```suggestion
          dayjs(row.getValue(columnId)).isSameOrBefore(
```

anujk14 (changes requested):

packages/raystack/table/FilteredChip.tsx
rsbh: Using functions inside the React component return statement is like an anti pattern.
We can create component like `SelectFilter`, `DateRangeFilter` etc
https://unicorn-utterances.com/posts/functions-are-killing-react-performance

packages/raystack/calendar/range-picker.tsx
rsbh: I added comments and simplified the logic here.

packages/raystack/table/FilteredChip.tsx
anujk14: Got it. Does it make sense to convert the inner logic to components instead of functions? [Like it is mentioned here in the eslint github repo](https://github.com/jsx-eslint/eslint-plugin-react/issues/578)?

So code becomes:
```
switch(valueType) {
  case filterValueTypeMap.select {
     <MySelectComponent prop1={} prop2={} />
  }
  case filterValueTypeMap.datePicker {
    <MyDatePickerComponent prop1={} prop2={} />
  }
}
```

PR #105, opened 2024-05-23
feat: add date and range picker component

rsbh:


packages/raystack/textfield/textfield.tsx
```diff
@@ -15,33 +15,53 @@ const textfield = cva(styles.textfield, {
...
 });
 
-type TextfieldProps = Omit<InputHTMLAttributes<HTMLInputElement>, "size"> &
+export type TextfieldProps = Omit<
+  InputHTMLAttributes<HTMLInputElement>,
+  "size"
+> &
   PropsWithChildren<VariantProps<typeof textfield>> & {
     leading?: React.ReactElement;
+    tailing?: React.ReactElement;
```
anujk14, line 36: Do we mean tailing, or trailing here? (like lead and trail)

PR #115, opened 2024-06-14
feat: DataTable now accepts onRowClick

singh-pk:


rsbh:
Add cursor pointer and add hover state styling

singh-pk:
> Add cursor pointer and add hover state styling

This is resolved

rsbh (approved):

PR #140, opened 2024-08-23
feat (IDE-555): Add infinite scroll to table component

paanSinghCoder:
IDE-555

1. Implement Infinite scroll in DataTable component. Table component accepts `onLoadMore` prop for infinite scroll.
2. Add bottom table loader.
3. Implement DataTable infinite scroll with dummy API in Example table.
4. Changed Example table layout.
5. Add Faker JS
6. Minor code refactor.

**Todo:**
1. Add support for filters.
2. Page number pagination.

**Testing:**
Tested on localhost with Example (www) project in Apsara.

infinite_scroll

PR #141, opened 2024-09-03
Fix/infinite scroll

paanSinghCoder:
Parent PR - [PR-140](https://github.com/raystack/apsara/pull/140)

- Add correct loading ui.
- Implement logic in effect hook as observer ref was not attaching to the last table row correctly bacause of async API call. 

**Why break down the logic into two useEffects?**
- Initialising intersection observer and attaching it to last row are dependent on seperate dependencies (onLoadMore and isLoading). In the previous useCallback approach, the init was triggered multiple times even when isLoading changed.

**First useEffect:**
Initialises the Intersection Observer (observerRef) when the onLoadMore function is passed in props. Threshold is set to 10% i.e. this will be triggered when 10% of the row has intersected.

**Second useEffect:**
Attaches the observer to the last row of the table, triggered whenever isLoading changes.

**Issues found:**
[here](https://github.com/raystack/frontier/pull/740/files#diff-253895be3895b3ec98796a7437de5ecc5414efca71db1a79c2ed35f24b6d6a5bL65) the global loader is conflicting with the infinite scroll. Functionality working fine.

**Testing:**
Manually tested with www/example and Frontier Admin UI

### DIFF
--- a/apps/www/examples/shield-ts/assets.tsx
+++ b/apps/www/examples/shield-ts/assets.tsx
@@ -114,7 +114,11 @@ export const Assets = () => {
   }, [isLoading, hasMoreData, page]);
 
   useEffect(() => {
-    setData(getData())
+    setIsLoading(true)
+    setTimeout(() => {
+      setData(getData())
+      setIsLoading(false)
+    }, 1000)
   }, [])
 
   return (
--- a/apps/www/examples/shield-ts/assets.tsx
+++ b/apps/www/examples/shield-ts/assets.tsx
@@ -114,11 +114,7 @@ export const Assets = () => {
   }, [isLoading, hasMoreData, page]);
 
   useEffect(() => {
-    setIsLoading(true)
-    setTimeout(() => {
-      setData(getData())
-      setIsLoading(false)
-    }, 1000)
+    loadMoreData()
   }, [])
 
   return (
--- a/packages/raystack/table/datatable.tsx
+++ b/packages/raystack/table/datatable.tsx
@@ -94,6 +94,37 @@ function DataTableRoot<TData, TValue>({
 
   const [tableState, setTableState] = useState<Partial<TableState>>({});
   const observerRef = useRef<IntersectionObserver | null>(null);
+  const lastRowRef = useRef<HTMLTableRowElement | null>(null);
+
+  useEffect(() => {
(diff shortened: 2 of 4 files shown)


apps/www/examples/shield-ts/assets.tsx
```diff
@@ -114,7 +114,11 @@ export const Assets = () => {
   }, [isLoading, hasMoreData, page]);
 
   useEffect(() => {
-    setData(getData())
+    setIsLoading(true)
```
rsbh, line 117: just call the `loadMoreData` here

packages/raystack/table/datatable.tsx
```diff
@@ -94,6 +94,37 @@ function DataTableRoot<TData, TValue>({
 
   const [tableState, setTableState] = useState<Partial<TableState>>({});
   const observerRef = useRef<IntersectionObserver | null>(null);
+  const lastRowRef = useRef<HTMLTableRowElement | null>(null);
+
+  useEffect(() => {
```
paanSinghCoder, line 99: Initialises the Intersection Observer (observerRef) when the onLoadMore function is passed in props. Threshold is set to 10% i.e. this will be triggered when 10% of the row has intersected.

packages/raystack/table/datatable.tsx
```diff
@@ -94,6 +94,37 @@ function DataTableRoot<TData, TValue>({
...
+        onLoadMore();
+      }
+    }, { threshold: 0.1 });
+
+    observerRef.current = observer;
+
+    return () => observer.disconnect();
+  }, [onLoadMore]);
+
+  useEffect(() => {
```
paanSinghCoder, line 113: Attaches the observer to the last row of the table, triggered whenever isLoading changes.

packages/raystack/table/datatable.tsx
```diff
@@ -94,6 +94,37 @@ function DataTableRoot<TData, TValue>({
...
+    const observer = observerRef.current;
+    const lastRow = lastRowRef.current;
+
+    if (observer && lastRow) {
```
paanSinghCoder, line 117: Detach the observer from the row if user has scrolled past it and there is a new last row. Attach to the new last row.

rsbh (approved):

PR #147, opened 2024-09-25
fix: DatePicker outside click causing stale date updates

singh-pk:
### Changelog

**Issue:** On clicking outside the date-picker, the date-picker was reverting back to the previous date

**Cause:** handleMouseDown is wrapped under useCallback so that document.removeEventListener('mouseup', handleMouseDown) receives the same instance of handleMouseDown as document.addEventListener('mouseup', handleMouseDown). But due to this the calendarVal inside handleMouseDown -> removeEventListener was getting stale value

**Screenshot:**

https://github.com/user-attachments/assets/1a61d745-06be-4b0f-bf0d-c166a62c3454

### DIFF
--- a/packages/raystack/calendar/date-picker.tsx
+++ b/packages/raystack/calendar/date-picker.tsx
@@ -42,6 +42,11 @@ export function DatePicker({
   const textFieldRef = useRef<HTMLInputElement | null>(null);
   const contentRef = useRef<HTMLDivElement | null>(null);
   const isInputFieldFocused = useRef(false);
+	const calendarValRef = useRef(value);
+
+	useEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/raystack/calendar/date-picker.tsx
```diff
@@ -56,14 +61,14 @@ export function DatePicker({
 
   function registerEventListeners() {
     isInputFieldFocused.current = true;
-    document.addEventListener('mouseup', handleMouseDown, true);
+    document.addEventListener('mouseup', handleMouseDown);
```
paanSinghCoder, line 64: Are these event listeners removed on component destroy phase?

paanSinghCoder (approved):

PR #172, opened 2024-11-07
Feat/v2 config

paanSinghCoder:
Setup rollup config for v1 components and styles.

Edit: Renamed all instances of `v2` to `v1`


packages/raystack/package.json
```diff
@@ -11,7 +11,11 @@
       "import": "./dist/index.js",
       "require": "./dist/index.cjs"
     },
-    "./index.css": "./dist/index.css"
+    "./v2": {
+      "import": "./dist/v2/index.js",
+      "require": "./dist/v2/index.cjs"
+    },
+    "./style.css": "./dist/style.css"
```
paanSinghCoder, line 18: Both v1 and v2 styles will be merged into a single file during build. This will not have conflicts as --rs- prefix is present for all v2 vars.

packages/raystack/package.json
```diff
@@ -27,6 +31,12 @@
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist && rm -rf react/dist",
     "test": "echo \"Error: no test specified\" && exit 0"
   },
+  "typesVersions": {
```
paanSinghCoder, line 34: Might not be required.

packages/raystack/rollup.config.mjs
```diff
@@ -5,49 +5,82 @@ import typescript from "@rollup/plugin-typescript";
...
-      sourcemap: true,
-      exports: "named",
+const createPlugins = () => [
+  nodeResolve(),
+  commonjs(),
+  postcss({
+    plugins: [
+      postcssImport(),
+    ],
+    extract: 'style.css',
```
paanSinghCoder, line 15: Kept the name same as previous styles file to keep the style file import same as before.

packages/raystack/package.json
paanSinghCoder, line 34: This might not be needed. Checking.

rsbh (approved):

packages/raystack/tsconfig.json
```diff
@@ -6,9 +6,10 @@
     "target": "esnext",
     "isolatedModules": false,
     "paths": {
-      "~/*": ["./*"]
+      "~/*": ["./*"],
+      "@raystack/apsara/v2": ["./v2/*"]
```
rsbh, line 10: why cant we use `~/v2` here instead of `@raystack/apsara/v2` ?

paanSinghCoder: `@raystack/apsara/v2` makes it more clear about where to import from.

rsbh: This is for internal code, why should someone import files from `@raystack/apsara/v2` inside the source of `@raystack/apsara` ?

paanSinghCoder: Changed to `~/v2/*`.

ravisuhag: @paanSinghCoder i think should use v1. since we have not launched stable version of apsara. and these changes will actually result in v1 since curent versions are still at 0.x cc @rsbh

paanSinghCoder: Makes sense! We started using the terms `v1` and `v2` from day one which brought us to this point. 

Moreover, In my opinion the new version might take some time to be fully stable. But it still makes more sense to call it v1 rather than v2.

paanSinghCoder: Renamed v2 to v1 everywhere. Merging.

PR #201, opened 2024-12-12
Feat/text area v1

paanSinghCoder:
Text area v1 component

### DIFF
--- a/packages/raystack/v1/components/textArea/textArea.tsx
+++ b/packages/raystack/v1/components/textArea/textArea.tsx
@@ -1,0 +1,43 @@
+import * as React from "react";
+import { HTMLAttributes, PropsWithChildren } from "react";
+import { VariantProps, cva, cx } from "class-variance-authority";
+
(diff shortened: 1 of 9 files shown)


packages/raystack/v1/components/textArea/textArea.tsx
```diff
@@ -0,0 +1,56 @@
...
+    error: {
+      true: styles.error
+    }
+  }
+});
+
+export interface TextAreaProps extends PropsWithChildren<VariantProps<typeof textArea>>, 
+  HTMLAttributes<HTMLTextAreaElement> {
+  label?: string;
+  isOptional?: boolean;
```
rsbh, line 19: why cant we use `required` props here?
we are using `required` in other components

paanSinghCoder: This was used in the figma design. I agree that required makes more sense and will be consistent with other component. Changing to required.

rsbh (approved):

PR #228, opened 2025-01-14
feat: Upgrade to React v19.0.0

paanSinghCoder:
**Changes:**

1. Update dependencies and peer dependencies to React v19.0.0.
2. Remove `elementRef` (deprecated) and use `componentRef`.
3. Remove `forwardRef` (deprecated). React v19 now supports passing `ref` directly as a prop thus forwardRef exclusively is not needed.
4. `defaultProps` is no more supported as ES6 default values is a better option.
6. Formatted code.
7. Minor fixes.

**Note: Old components are untouched.**

### DIFF
--- a/packages/raystack/v1/components/button/button.tsx
+++ b/packages/raystack/v1/components/button/button.tsx
@@ -1,6 +1,6 @@
 import { Slot } from "@radix-ui/react-slot";
 import { cva, type VariantProps } from "class-variance-authority";
-import { ButtonHTMLAttributes, forwardRef, PropsWithChildren, ReactNode } from "react";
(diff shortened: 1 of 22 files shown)


packages/raystack/v1/components/button/button.tsx
```diff
@@ -41,7 +41,10 @@ type ButtonProps = PropsWithChildren<VariantProps<typeof button>> &
     width?: string | number
   };
 
-export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
+export const Button = forwardRef<
```
ravisuhag, line 44: @paanSinghCoder Do we still need the forwardRef function in 19?

paanSinghCoder: Removed forwardRef as deprecated and directly passed ref as prop.

PR #386, opened 2025-05-06
feat: Support for `as` in Sidebar.Item

rohanchkrabrty:
## Description

This PR introduces as support for sidebar items

## Type of Change

- [ ] Documentation update
- [x] Improvement (Improvements to existing code)
- [ ] Other (please specify)

## How Has This Been Tested?

Manual


packages/raystack/v1/components/sidebar/sidebar.tsx
```diff
@@ -38,6 +42,7 @@ interface SidebarItemProps extends ComponentPropsWithoutRef<"a"> {
   icon: ReactNode;
   active?: boolean;
   disabled?: boolean;
+  as?: ReactElement;
```
paanSinghCoder, line 45: Let's update the docs as well.

paanSinghCoder (changes requested):

packages/raystack/v1/components/sidebar/sidebar.tsx
rohanchkrabrty: Yes

packages/raystack/v1/components/sidebar/sidebar.tsx
```diff
@@ -149,44 +159,52 @@ const SidebarFooter = forwardRef<HTMLDivElement, SidebarFooterProps>(
...
-      {...props}
-    >
+      {...props}>
       {children}
     </div>
-  )
+  ),
 );
 
 const SidebarItem = forwardRef<HTMLAnchorElement, SidebarItemProps>(
```
paanSinghCoder, line 168: `as` can render anything (and not just the anchor tag), thus it is restrictive to only have `HTMLAnchorElement` with `forwardRef`. A generic type (`<T ...`) would be more appropriate here which matches with the prop type passed.

paanSinghCoder (changes requested):

packages/raystack/v1/components/sidebar/sidebar.tsx
rohanchkrabrty: It's meant to be an AnchorRef itself as it's a Link element, and as a default it's anchor. So the ref makes sense.

As I mentioned previously components that support `asChild` is no different from this. We don't do generic ref in that case so it shouldn't be here

paanSinghCoder (approved):

PR #737, opened 2026-04-14
fix(switch): replace hardcoded pixel widths with design token expressions

rohanchkrabrty:
## Summary
- Replace hardcoded `34px` and `26px` width values in Switch component CSS with `calc()` expressions using existing `--rs-*` spacing tokens
- Large switch: `calc(var(--rs-space-5) * 2 + var(--rs-space-1))` (= 34px)
- Small switch: `calc(var(--rs-space-4) * 2 + var(--rs-space-1))` (= 26px)

## Test plan
- [x] All 1619 existing tests pass
- [x] Visual output unchanged (values are mathematically equivalent)
- [ ] Verify Switch renders correctly in browser at both sizes

### DIFF
--- a/packages/raystack/components/switch/switch.module.css
+++ b/packages/raystack/components/switch/switch.module.css
@@ -12,12 +12,12 @@
 }
 
 .switch.large {
-  width: 34px;
+  width: calc(var(--rs-space-5) * 2 + var(--rs-space-1));
   height: var(--rs-space-6);
 }
 
 .switch.small {
-  width: 26px;
+  width: calc(var(--rs-space-4) * 2 + var(--rs-space-1));
   height: var(--rs-space-5);
 }
 


paanSinghCoder:
I think it will be a better idea to either add new CSS variable for the missing px values or use existing variable instead of composing a new value. We can discuss this with design.

Composing a new value might introduce unexpected changes. For example, if `calc(var(--rs-space-5) * 2 + var(--rs-space-1))` `--rs-space-5` changes from `16px` to `18px` in future it will change the large switch from `34px` to `38px`.

ravisuhag:
Yes, I think let's not use combination cals. Either we should standarize it or introduce vars. Or for such one off scenarios, leave it hardcode.

rohanchkrabrty:
@ravisuhag @paanSinghCoder Makes sense. Leaving it hardcoded for now due to one off scenario. Closing the PR

PR #738, opened 2026-04-14
fix: textarea overflow, default rows, and size/variant API

### LINKED ISSUE #645: [TextArea] Fix label-textarea association and add aria-invalid/aria-describedby
## Summary

TextArea is a custom HTML textarea wrapper with label, helper text, and error state support. Has critical accessibility gaps around label association and ARIA attributes.

### 1. Add `htmlFor`/`id` association between label and textarea
The `<label>` element has no `htmlFor` attribute linked to the textarea. Screen readers cannot associate the label with the control. Should auto-generate an ID if not provided.

### 2. Add `aria-invalid` for error state
When `error={true}`, should set `aria-invalid="true"` on the textarea element.

### 3. Add `aria-describedby` for helper/error text
Helper text and error messages are not linked to the textarea. Should use `aria-describedby` for semantic connection.

### 4. Fix `overflow: hidden` on textarea
CSS has `overflow: hidden` which clips content. Should be `overflow: auto` or paired with explicit resize control.

### 5. Replace hardcoded icon dimensions
Helper icon has `width: 12px; height: 12px` hardcoded instead of design tokens.

### 6. Add `minRows`/`maxRows` props
No height control beyond fixed CSS. Should support row-based height constraints.

### 7. Align API with InputField
InputField has `size` and `variant` CVA variants. TextArea doesn't follow the same pattern.

## Files
- `packages/raystack/raystack/components/text-area/`

rohanchkrabrty:
## Summary
- **Fix overflow: hidden** on textarea CSS -- changed to `overflow: auto` so users can scroll when content exceeds the visible area
- **Replace CSS min-height with native `rows` attribute** -- removed hardcoded `min-height: var(--rs-space-13)` and set `rows={3}` as default, allowing users to override with `<TextArea rows={6} />`
- **Add `size` and `variant` CVA variants** to align TextArea API with InputField -- supports `size="small" | "large"` and `variant="default" | "borderless"` with matching defaults (`size: 'large'`, `variant: 'default'`)

## Test plan
- [x] All 23 TextArea tests pass (including 8 new tests for rows, sizes, and variants)
- [x] Full test suite passes (1625 tests across 70 files)
- [x] No new TypeScript errors introduced
- [x] Verify textarea renders with default 3 rows height
- [x] Verify `<TextArea rows={6} />` renders taller than default
- [x] Verify `size="small"` applies smaller padding
- [x] Verify `variant="borderless"` removes borders
- [x] Verify scrollbar appears when content overflows

Closes #645

### DIFF
--- a/apps/www/src/content/docs/components/textarea/demo.ts
+++ b/apps/www/src/content/docs/components/textarea/demo.ts
@@ -23,6 +23,11 @@ export const playground = {
       options: ['default', 'borderless'],
       defaultValue: 'default'
     },
+    rows: {
+      type: 'number',
+      defaultValue: 3,
+      min: 1
+    },
     width: {
       type: 'text',
       defaultValue: '400px'
--- a/apps/www/src/content/docs/components/textarea/demo.ts
+++ b/apps/www/src/content/docs/components/textarea/demo.ts
@@ -13,6 +13,16 @@ export const playground = {
       type: 'checkbox',
       defaultValue: false
     },
+    size: {
+      type: 'select',
+      options: ['large', 'small'],
+      defaultValue: 'large'
+    },
+    variant: {
+      type: 'select',
+      options: ['default', 'borderless'],
+      defaultValue: 'default'
+    },
     width: {
       type: 'text',
       defaultValue: '400px'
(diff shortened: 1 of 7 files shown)


packages/raystack/components/text-area/text-area.module.css
```diff
@@ -61,3 +60,34 @@
 .textarea[data-invalid]:focus {
   border-color: var(--rs-color-border-danger-emphasis-hover);
 }
+
+/* Size variants */
+.size-large {
```
paanSinghCoder, line 65: `.size-large` is not required as its the default `size` and these three properties are already defined in `.textarea` class.

packages/raystack/components/text-area/text-area.module.css
```diff
@@ -5,15 +5,14 @@
   outline: none;
   height: auto;
   width: 100%;
-  min-height: var(--rs-space-13);
```
paanSinghCoder, line 8: Removing `min-height` might introduce unexpected height shift in the client. We should mention this in `v1-migration-guide`. 

Example: If someone sets `rows={1}`, it will render a shorter `textarea` than before.

apps/www/src/content/docs/components/textarea/demo.ts
```diff
@@ -13,6 +13,16 @@ export const playground = {
...
+    size: {
+      type: 'select',
+      options: ['large', 'small'],
+      defaultValue: 'large'
+    },
+    variant: {
+      type: 'select',
+      options: ['default', 'borderless'],
+      defaultValue: 'default'
+    },
```
paanSinghCoder, line 25: Lets add rows prop control as well in the playground.

packages/raystack/components/text-area/text-area.module.css
```diff
@@ -61,3 +60,34 @@
...
+}
+
+.variant-borderless:hover {
+  border-color: transparent;
+}
+
+.variant-borderless:focus:not(:disabled) {
+  border-color: transparent;
+  outline: none;
+}
```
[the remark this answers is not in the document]
paanSinghCoder: We have focus ring related improvements planned. Will be picked later.

paanSinghCoder (approved):

PR #739, opened 2026-04-14
feat(radio): add size variants, orientation prop, and fix CSS issues

### LINKED ISSUE #630: [Radio] Add focus-visible styles, size variants, and export TypeScript types
## Summary

Radio wraps Base UI's Radio/RadioGroup primitives. Missing keyboard focus styles (WCAG violation), size variants, and TypeScript type exports.

### 1. Add `:focus-visible` styles (WCAG 2.4.7)
No focus indicator for keyboard navigation. Critical accessibility gap — Button and other components have this.

### 2. Add `'use client'` directive
Missing directive for Server Component compatibility. Checkbox has it; Radio doesn't.

### 3. Export `RadioProps` and `RadioGroupProps` TypeScript interfaces
No exported type definitions. Switch exports `SwitchProps` — Radio should follow the same pattern.

### 4. Add size variants
Radio uses fixed `var(--rs-space-5)` sizing. Switch has `size` prop with 'small' | 'large' variants. Radio should match.

### 5. Add `orientation` prop to Radio.Group
No way to control layout direction (vertical/horizontal). Base UI supports this but it's not exposed.

### 6. Fix CSS issues
- Duplicate `display` property (line 7: `display: flex`, line 15: `display: inline-flex`)
- Indicator uses `var(--rs-radius-3)` for width/height — semantically a radius variable, not a size

### 7. Align disabled state styling with Checkbox
Radio uses different colors for disabled state; Checkbox uses opacity approach. Should be consistent.

## Files
- `packages/raystack/raystack/components/radio/`

rohanchkrabrty:
## Summary
- **Size variants**: Add `size` prop (`small` | `large`) to Radio using CVA, following the same pattern as the Switch component. Default is `large` (preserving current 16px size). Small variant uses `--rs-space-4` (12px).
- **Orientation prop**: Add `orientation` prop (`vertical` | `horizontal`) to `Radio.Group` for controlling layout direction. Default is `vertical`. Implemented as a custom Apsara feature since Base UI's RadioGroup does not natively support orientation.
- **Fix indicator token**: Replace `--rs-radius-3` (a border-radius token) with `--rs-space-2` (a spacing token) for the indicator dot width/height, fixing semantic token misuse.
- **CSS audit**: Rename `.radio` to `.group`, add clarifying CSS comments for the display property usage on `.group` (flex for layout) and `.radioitem` (inline-flex for centering indicator), and remove fixed dimensions from `.radioitem` base class (moved to size variants).
- **Proportional indicator scaling**: The indicator dot scales down for the small size variant using `--rs-space-1`.

Closes #630

## Test plan
- [x] All 22 radio tests pass (including 7 new tests for sizes and orientation)
- [x] Full test suite passes (1626 tests across 70 files)
- [x] No new type errors introduced (verified with `tsc --noEmit`)
- [ ] Visually verify radio indicator dot renders correctly at both sizes
- [ ] Verify horizontal/vertical orientation layout in Storybook

### DIFF
--- a/apps/www/src/content/docs/components/radio/props.ts
+++ b/apps/www/src/content/docs/components/radio/props.ts
@@ -17,6 +17,12 @@ export interface RadioGroupProps {
    */
   orientation?: 'vertical' | 'horizontal';
 
+  /**
+   * The default size applied to all radio items in the group. Can be overridden per item.
+   * @default "large"
+   */
+  size?: 'large' | 'small';
+
   /** The name of the radio group when submitted as a form field. */
   name?: string;
 
@@ -42,8 +48,8 @@ export interface RadioProps {
   disabled?: boolean;
 
   /**
-   * The size of the radio button.
-   * @default "large"
+   * The size of the radio button. Overrides the `size` set on `Radio.Group`.
+   * If neither is set, defaults to `"large"` via the group.
    */
   size?: 'large' | 'small';
 
--- a/apps/www/src/content/docs/components/radio/props.ts
+++ b/apps/www/src/content/docs/components/radio/props.ts
@@ -11,6 +11,12 @@ export interface RadioGroupProps {
   /** When true, prevents user interaction with the radio group. */
   disabled?: boolean;
 
+  /**
+   * The layout orientation of the radio group.
+   * @default "vertical"
+   */
+  orientation?: 'vertical' | 'horizontal';
+
   /** The name of the radio group when submitted as a form field. */
   name?: string;
 
@@ -34,6 +40,12 @@ export interface RadioProps {
 
   /** When true, prevents user interaction with this radio item. */
   disabled?: boolean;
+
+  /**
+   * The size of the radio button.
+   * @default "large"
+   */
+  size?: 'large' | 'small';
 
   /** The unique identifier for the radio item. */
   id?: string;
--- a/packages/raystack/components/radio/radio.tsx
+++ b/packages/raystack/components/radio/radio.tsx
@@ -1,21 +1,42 @@
 import { Radio as RadioPrimitive } from '@base-ui/react/radio';
 import { RadioGroup as RadioGroupPrimitive } from '@base-ui/react/radio-group';
-import { cx } from 'class-variance-authority';
+import { cva, cx, type VariantProps } from 'class-variance-authority';
 import { useFieldContext } from '../field';
 
 import styles from './radio.module.css';
 
+const radioVariants = cva(styles.radioitem, {
+  variants: {
+    size: {
+      small: styles.small,
+      large: styles.large
+    }
+  },
+  defaultVariants: {
+    size: 'large'
+  }
+});
+
+interface RadioGroupProps extends RadioGroupPrimitive.Props {
+  orientation?: 'vertical' | 'horizontal';
+}
+
 function RadioGroup({
   className,
+  orientation = 'vertical',
   required,
   ...props
-}: RadioGroupPrimitive.Props) {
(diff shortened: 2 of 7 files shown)


packages/raystack/components/radio/radio.tsx
```diff
@@ -1,21 +1,42 @@
...
   ...props
-}: RadioGroupPrimitive.Props) {
+}: RadioGroupProps) {
   const fieldContext = useFieldContext();
   const resolvedRequired = required ?? fieldContext?.required;
 
   return (
     <RadioGroupPrimitive
-      className={cx(styles.radio, className)}
+      className={cx(
```
paanSinghCoder, line 35: I think based on the `orientation` prop we can set the `aria-orientation` as well.

[aria-orientation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-orientation) Even though these guys have not listed `Radio` in their docs but I strongly think we can add it. Thoughts?

packages/raystack/components/radio/radio.tsx
```diff
@@ -24,9 +45,16 @@ function RadioGroup({
 
 RadioGroup.displayName = 'Radio.Group';
 
-function RadioItem({ className, ...props }: RadioPrimitive.Root.Props) {
+interface RadioItemProps
+  extends RadioPrimitive.Root.Props,
+    VariantProps<typeof radioVariants> {}
+
+function RadioItem({ className, size, ...props }: RadioItemProps) {
```
paanSinghCoder, line 52: I think we should move the `size` prop to `RadioGroup` as user will have to add it 5 times if there are 5 RadioItems. size is almost always a Group thing.

packages/raystack/components/radio/radio.tsx
```diff
@@ -1,21 +1,42 @@
...
+}: RadioGroupProps) {
   const fieldContext = useFieldContext();
   const resolvedRequired = required ?? fieldContext?.required;
 
   return (
     <RadioGroupPrimitive
-      className={cx(styles.radio, className)}
+      className={cx(
+        styles.group,
+        orientation === 'horizontal' && styles['group-horizontal'],
```
paanSinghCoder, line 37: `orientation === 'horizontal' && styles['group-horizontal'],`

Can be changed to something like `className={radioGroupVariants({ orientation, className })}`, similar to `radioItem` on line 56.

apps/www/src/content/docs/components/radio/props.ts
```diff
@@ -11,6 +11,12 @@ export interface RadioGroupProps {
   /** When true, prevents user interaction with the radio group. */
   disabled?: boolean;
 
+  /**
+   * The layout orientation of the radio group.
+   * @default "vertical"
```
paanSinghCoder, line 16: We are using `@default` in some components props and `@defaultValue` in other. We can standardise this later.

rohanchkrabrty: Can be picked later as cross cutting improvement

packages/raystack/components/radio/radio.tsx
rohanchkrabrty: It's not a group thing, but we can add a prop to group to prevent code duplication. Individual prop will override group size

paanSinghCoder: Yes, by group thing I meant size is something which is usually the same for all items and makes sense to have a common prop in group that controls all item sizes at once.

paanSinghCoder (approved):

PR #742, opened 2026-04-14
fix(input-field): fix pointer-events, disable chips when disabled, standardize CSS naming

rohanchkrabrty:
## Summary
- **Fix duplicate pointer-events:** Removed contradictory `pointer-events: none` declarations from `.leading-icon` and `.trailing-icon`, keeping only `pointer-events: auto` so icons remain clickable as intended
- **Disable chips when input is disabled:** Added `disabled` prop to `Chip` component with `data-disabled` attribute and CSS styling (`pointer-events: none; opacity: 0.5`). InputField now passes its disabled state to chips, preventing dismiss button rendering and interaction
- **Standardize CSS naming:** Renamed `.inputWrapper` to `.input-wrapper` to match the kebab-case convention used by the other 17 CSS classes in the module

## Test plan
- [x] All 44 existing InputField tests pass with updated CSS class references
- [x] All 23 existing Chip tests continue to pass
- [x] 3 new InputField tests: disabled chips hide dismiss button, set data-disabled, and remain interactive when not disabled
- [x] 3 new Chip tests: data-disabled attribute present/absent, onClick suppressed when disabled
- [x] TypeScript compiles with no new errors
- [x] Biome formatting passes

### DIFF
--- a/packages/raystack/components/input-field/input-field.module.css
+++ b/packages/raystack/components/input-field/input-field.module.css
@@ -62,7 +62,6 @@
   height: var(--rs-space-5);
   color: var(--rs-color-foreground-base-secondary);
   margin-right: var(--rs-space-3);
-  pointer-events: auto;
 }
 
 .input-field {
--- a/packages/raystack/components/input-field/input-field.module.css
+++ b/packages/raystack/components/input-field/input-field.module.css
@@ -1,6 +1,6 @@
 /* Note: If making changes here, possibly also need to make changes in text-area.module.css for design consistency. */
 
-.inputWrapper {
+.input-wrapper {
   display: flex;
   align-items: center;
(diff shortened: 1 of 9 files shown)


rsbh (approved):

packages/raystack/components/input-field/input-field.module.css
```diff
@@ -50,7 +50,6 @@
   width: var(--rs-space-5);
   height: var(--rs-space-5);
   color: var(--rs-color-foreground-base-secondary);
-  pointer-events: none;
   margin-left: var(--rs-space-3);
   pointer-events: auto;
```
paanSinghCoder, line 54: `Pointer-events: auto;` is the default property. We can remove it.

PR #745, opened 2026-04-20
feat(floating-actions): add FloatingActions component

paanSinghCoder:
## Summary
- New `FloatingActions` component — a contextual action bar (bulk-action toolbar, row-hover actions) built on Base UI's [Toolbar](https://base-ui.com/react/components/toolbar) primitive, so it inherits roving focus, arrow-key navigation, and the `disabled` cascade.
- Two layout modes via `variant`: `floating` (default — `position: fixed`, anchored via `side`/`align`, defaults to bottom-center) and `inline` (in-flow). No positioning CSS at the call site.
- Ships `FloatingActions.Group` and `FloatingActions.Separator`. Composes with existing `Chip`, `Button`, `IconButton`. Docs cover variants, grouped actions, icon-only, and a scrolling-context demo; Vitest suite + playground wired up.

## Usage
```tsx
<FloatingActions aria-label="Selection actions">
  <Chip isDismissible leadingIcon={<CheckCircledIcon />}>2 selected</Chip>
  <FloatingActions.Separator />
  <Button variant="outline" color="neutral" size="small">Move to</Button>
</FloatingActions>
```

## Test plan
- [ ] `pnpm --filter @raystack/apsara test components/floating-actions`
- [ ] Docs site renders the variants, grouped actions, icon-only, and scrolling-context demos

### DIFF
--- a/apps/www/src/content/docs/components/floating-actions/index.mdx
+++ b/apps/www/src/content/docs/components/floating-actions/index.mdx
@@ -5,16 +5,16 @@ source: packages/raystack/components/floating-actions
 tag: new
 ---
 
-import { preview, floatingDemo, bulkActionsDemo, iconOnlyDemo } from "./demo.ts";
+import { preview, floatingDemo, bulkActionsDemo, iconOnlyDemo, groupDemo, scrollingDemo } from "./demo.ts";
 
 <Demo data={preview} />
 
 ## Anatomy
 
-Import and assemble the component along with other primitives. `FloatingActions` provides the container (built on top of [`Separator`](/docs/components/separator) for the divider); everything inside is freely composable.
+Import and assemble the component:
 
 ```tsx
-import { FloatingActions, Chip, Button } from '@raystack/apsara'
+import { FloatingActions, Chip, Button } from "@raystack/apsara";
 
 <FloatingActions aria-label="Selection actions">
   <Chip isDismissible>2 selected</Chip>
@@ -26,15 +26,23 @@ import { FloatingActions, Chip, Button } from '@raystack/apsara'
 
 ## API Reference
 
+Built on top of Base UI's [Toolbar](https://base-ui.com/react/components/toolbar) primitive, so keyboard navigation, roving focus, and the `role="toolbar"` semantics come from the primitive.
+
 ### Root
 
-The container. Renders as a `role="toolbar"` element by default and pins itself to the viewport (`variant="floating"`).
+The container. Pins itself to the viewport by default (`variant="floating"`) and forwards the underlying Toolbar primitive props (`disabled`, `orientation`, `loopFocus`, plus all `<div>` attributes).
 
 <auto-type-table path="./props.ts" name="FloatingActionsProps" />
 
+### Group
+
+Cluster related items so that the toolbar treats them as a single navigation unit. Useful when you have a few destructive actions you want to disable together (`disabled` cascades to every focusable child).
+
+<auto-type-table path="./props.ts" name="FloatingActionsGroupProps" />
+
 ### Separator
 
-A vertical divider sized to the bar's content.
+A vertical divider sized to the bar's content. Built on top of [`Separator`](/docs/components/separator).
 
 <auto-type-table path="./props.ts" name="FloatingActionsSeparatorProps" />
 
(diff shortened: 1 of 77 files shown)


packages/raystack/components/floating-actions/floating-actions.tsx
```diff
@@ -0,0 +1,38 @@
...
+    orientation='vertical'
+    className={cx(styles.separator, className)}
+    {...props}
+  />
+);
+FloatingActionsSeparator.displayName = 'FloatingActions.Separator';
+
+export const FloatingActions = Object.assign(FloatingActionsRoot, {
+  Separator: FloatingActionsSeparator
+});
```
rohanchkrabrty: Split the object assign into a separate file or do it in index.ts

This pattern with client directive creates some weird client/server issues

packages/raystack/components/floating-actions/index.tsx
rohanchkrabrty, line 1: File name should be `index.ts` for barrel export

packages/raystack/components/floating-actions/floating-actions.tsx
```diff
@@ -0,0 +1,38 @@
...
+import styles from './floating-actions.module.css';
+
+export interface FloatingActionsProps extends ComponentProps<'div'> {}
+
+const FloatingActionsRoot = ({
+  className,
+  role = 'toolbar',
+  ...props
+}: FloatingActionsProps) => (
+  <div role={role} className={cx(styles.root, className)} {...props} />
```
rohanchkrabrty: Let' use Toolbar Primitive from Base UI.

Also let's expose a `Group` subcomponent

apps/www/src/content/docs/components/floating-actions/index.mdx
```diff
@@ -0,0 +1,59 @@
...
+tag: new
+---
+
+import { preview, bulkActionsDemo, iconOnlyDemo } from "./demo.ts";
+
+<Demo data={preview} />
+
+## Anatomy
+
+Compose the bar from existing Apsara primitives. `FloatingActions` provides the container and a matching vertical separator; everything inside is freely composable.
```
rohanchkrabrty: This should be a simpler line like -  Import and assemble the component along with other primtives

Follows other docs pattern

apps/www/src/content/docs/components/floating-actions/props.ts
```diff
@@ -0,0 +1,18 @@
...
+  role?: string;
+
+  /** Additional CSS class names. */
+  className?: string;
+
+  /** The contents of the floating bar. */
+  children?: React.ReactNode;
+}
+
+export interface FloatingActionsSeparatorProps {
```
rohanchkrabrty, line 37: All Separator props should be documented especially `color`

apps/www/src/content/docs/components/floating-actions/index.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+### Root
+
+The floating container. Renders as a `role="toolbar"` element by default.
+
+<auto-type-table path="./props.ts" name="FloatingActionsProps" />
+
+### Separator
+
+A vertical divider sized to the bar's content.
```
rohanchkrabrty: Mention that it's built on top of Separator component

apps/www/src/content/docs/components/floating-actions/index.mdx
rohanchkrabrty, line 1: There is no demo where the component is seen in a scrolling context.

rohanchkrabrty (changes requested):
1. IMO the `FloatingActions` component should have position agnostic props, like `side` and `align`.

There can be two variants -  `inline` and `floating`(default), `inline` will be the current one. But in `floating` the `FloatingAction` will be actually floating using the side and align position props.

Right now consumer has to write a a block of css just to use this component. The name also is `FloatingActions` but nowhere it supports that functionality, so it's misleading. 

2. I think it should also support `orientation` for cases where the component is used on the left or right. But this is can be picked up later too

packages/raystack/components/floating-actions/floating-actions.tsx
paanSinghCoder: Moved to index.tsx keeping same export pattern as other components.

packages/raystack/components/floating-actions/index.tsx
paanSinghCoder: Done

packages/raystack/components/floating-actions/floating-actions.tsx
paanSinghCoder: Done. Using the Base UI primitive. Also exposed the Group.

apps/www/src/content/docs/components/floating-actions/index.mdx
paanSinghCoder: Done

apps/www/src/content/docs/components/floating-actions/index.mdx
paanSinghCoder: Added to the docs.

apps/www/src/content/docs/components/floating-actions/index.mdx
paanSinghCoder: Added.

rohanchkrabrty (approved):

PR #747, opened 2026-04-21
docs(floating-actions): add row-selection pattern with DataTable

paanSinghCoder:
Screenshot 2026-04-21 at 9 13 53 AM

## Summary
- Adds a "Row selection" section to DataTable docs showing how to integrate `FloatingActions` with TanStack row selection (leading checkbox column + `useDataTable()` + dismissible count chip wired to `table.resetRowSelection()`).
- Ships a live `DataTableSelectionDemo` registered in the Demo scope so the pattern is visible in the docs, not just copy-pasted code.
- Cross-reference added in the `FloatingActions` docs pointing at the new section.
- Small DataTable fix: horizontal padding on `.toolbar` so filter/display controls align with cell content.

## Test plan
- [ ] Visit `/components/datatable` → scroll to "Row selection" — live demo renders, clicking checkboxes updates the "N selected" chip; dismissing clears selection.
- [ ] Visit `/components/floating-actions` → see cross-reference at "Use with DataTable".
- [ ] Confirm toolbar padding shift looks fine across other DataTable demos.


apps/www/src/content/docs/components/datatable/demo.ts
```diff
@@ -4,3 +4,8 @@ export const preview = {
   type: 'code',
   code: `<DataTableDemo />`
 };
+
+export const rowSelectionDemo = {
+  type: 'code',
+  code: `<DataTableSelectionDemo />`
```
rohanchkrabrty: we can use `codePreview` prop in the demo here to show the code directly in Markdown. [Example here](https://github.com/raystack/apsara/blob/main/apps/www/src/content/docs/components/menu/demo.ts#L267)

That way we don't need to show the codeblock in `.mdx` file

rohanchkrabrty (changes requested):

rohanchkrabrty:
@paanSinghCoder Also the selection state should be handled internally from the DataTable. Right now the user is supposed to handle the selection model, which is not ideal. Let's think this through separately

cc @ravisuhag

apps/www/src/content/docs/components/datatable/demo.ts
paanSinghCoder: Done

rohanchkrabrty (approved):

PR #748, opened 2026-04-21
feat: migrate command to Base UI

rohanchkrabrty:
## Summary

- Migrated Command component from `cmdk` to the Base UI Autocomplete primitive with built-in auto-search filtering
- Split monolithic `command.tsx` into dedicated subcomponent files (`command-root`, `command-input`, `command-list`, `command-item`, `command-dialog`, `command-empty`, `command-misc`)
- Added new subcomponents: `Panel`, `Footer`, `Label`, `Collection`, and a composable `Dialog` API (`Dialog.Trigger`, `Dialog.Content`, `Dialog.Close`, `Dialog.Title`, `Dialog.Description`)
- Expanded tests, documentation (API reference, usage examples), demos, and playground examples for the new API surface
- Updated CSS module with styles for all new subcomponents and refined existing styles

### DIFF
--- a/apps/www/src/components/docs/search.tsx
+++ b/apps/www/src/components/docs/search.tsx
@@ -158,7 +158,7 @@ export default function DocsSearch({ pageTree }: { pageTree: Root }) {
               </IconButton>
             )}
           </Flex>
-          <Command.List className={styles.searchList}>
+          <Command.Content className={styles.searchList}>
             <Command.Empty>
               <EmptyState
                 variant='empty1'
@@ -214,7 +214,7 @@ export default function DocsSearch({ pageTree }: { pageTree: Root }) {
                 {index < defaultItems.length - 1 && <Command.Separator />}
               </div>
             ))}
-          </Command.List>
+          </Command.Content>
         </Command>
       </Dialog.Content>
     </Dialog>
--- a/apps/www/src/components/playground/command-examples.tsx
+++ b/apps/www/src/components/playground/command-examples.tsx
@@ -24,54 +24,35 @@ export function CommandExamples() {
         <Text>Inline command menu</Text>
         <Flex style={{ width: 420 }}>
           <Command>
-            <Command.Panel>
-              <Command.Input placeholder='Type a command or search...' />
-              <Command.List>
-                <Command.Empty>No results found.</Command.Empty>
-                <Command.Group>
-                  <Command.Label>Suggestions</Command.Label>
-                  <Command.Item>Calendar</Command.Item>
-                  <Command.Item>Search Emoji</Command.Item>
-                  <Command.Item>Calculator</Command.Item>
-                </Command.Group>
-                <Command.Separator />
-                <Command.Group>
-                  <Command.Label>Settings</Command.Label>
-                  <Command.Item>
-                    Profile <Command.Shortcut>⌘P</Command.Shortcut>
-                  </Command.Item>
-                  <Command.Item>
-                    Billing <Command.Shortcut>⌘B</Command.Shortcut>
-                  </Command.Item>
-                  <Command.Item>
-                    Settings <Command.Shortcut>⌘S</Command.Shortcut>
-                  </Command.Item>
-                </Command.Group>
-              </Command.List>
-              <Command.Footer>
-                <span>
-                  Press <Command.Shortcut>↵</Command.Shortcut> to select
-                </span>
-                <span>
-                  <Command.Shortcut>ESC</Command.Shortcut> to close
-                </span>
-              </Command.Footer>
-            </Command.Panel>
+            <Command.Input placeholder='Type a command or search...' />
+            <Command.Content>
+              <Command.Empty>No results found.</Command.Empty>
+              <Command.Group>
+                <Command.Label>Suggestions</Command.Label>
+                <Command.Item>Calendar</Command.Item>
+                <Command.Item>Search Emoji</Command.Item>
+                <Command.Item>Calculator</Command.Item>
+              </Command.Group>
+              <Command.Separator />
+              <Command.Group>
+                <Command.Label>Settings</Command.Label>
+                <Command.Item>Profile</Command.Item>
+                <Command.Item>Billing</Command.Item>
+                <Command.Item>Settings</Command.Item>
+              </Command.Group>
+            </Command.Content>
           </Command>
         </Flex>
 
-        <Text>
-          Dialog — press <Command.Shortcut>⌘K</Command.Shortcut> or click the
-          button
-        </Text>
+        <Text>Dialog — press ⌘K or click the button</Text>
         <Command.Dialog open={open} onOpenChange={setOpen}>
-          <Command.Dialog.Trigger render={<Button variant='outline' />}>
+          <Command.DialogTrigger render={<Button variant='outline' />}>
             Open Command Menu
-          </Command.Dialog.Trigger>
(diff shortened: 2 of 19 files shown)


apps/www/src/components/docs/search.tsx
```diff
@@ -158,7 +158,7 @@ export default function DocsSearch({ pageTree }: { pageTree: Root }) {
               </IconButton>
             )}
           </Flex>
-          <Command.List className={styles.searchList}>
+          <Command.Content className={styles.searchList}>
```
paanSinghCoder, line 161: Changing List to Content is a breaking change. Lets add it to the migration doc.

apps/www/src/components/playground/command-examples.tsx
```diff
@@ -1,29 +1,75 @@
...
-        </Command>
+      <Flex direction='column' gap='large'>
+        <Text>Inline command menu</Text>
+        <Flex style={{ width: 420 }}>
+          <Command>
+            <Command.Input placeholder='Type a command or search...' />
+            <Command.Content>
+              <Command.Empty>No results found.</Command.Empty>
+              <Command.Group>
+                <Command.Label>Suggestions</Command.Label>
```
paanSinghCoder, line 31: Changing `Command.Group heading=""` to `<Command.Group><Command.Label>` is a breaking change. lets add it to migration doc.

apps/www/src/content/docs/components/command/index.mdx
```diff
@@ -1,30 +1,162 @@
...
+
+### Dialog
+
+`Command.Dialog` is an alias for the Base UI Dialog root. Pair it with `Command.DialogTrigger` and `Command.DialogContent` to render the command menu in a centered modal.
+
+<auto-type-table path="./props.ts" name="CommandDialogProps" />
+
+#### DialogContent
+
+Renders the portal, backdrop, viewport, and popup for the command palette. The backdrop is not blurred by default — pass `overlay={{ blur: true }}` to enable it. Unlike `Dialog.Content`, there is no close button; users dismiss the palette via `Escape` or by clicking the backdrop.
```
paanSinghCoder, line 112: Why is `overlay={{ blur: true }}` prop structure like this?

Moreover, in the recent past we have been pushing back on using styling as props and pass it as className instead. overlay is purely a styling thing. I think we can decide on keeping this purely a className rather than passing as prop. We can simply expose `backdropClassName`.

packages/raystack/components/command/command-input.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export const CommandInput = ({
+  ref,
+  leadingIcon,
+  autoFocus = true,
+  size = 'large',
+  placeholder = 'Search...',
+  ...props
+}: CommandInputProps) => {
```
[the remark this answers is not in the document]
paanSinghCoder: I think its the correct behaviour. Most of the user want auto focus on the input thus makes sense to keep it true by default. THere is always an option to toggle it to false.

packages/raystack/components/command/command-misc.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const keys =
+    typeof children === 'string'
+      ? children.trim().split(/\s+/).filter(Boolean)
+      : Array.isArray(children)
+        ? children
+        : [children];
+
+  return (
+    <span className={cx(styles.shortcut, className)} {...props}>
+      {keys.map((key, index) => (
```
paanSinghCoder, line 79: Any way we can have an option to pass a stable index? If not lets leave it to index.

packages/raystack/components/command/command-item.tsx
```diff
@@ -0,0 +1,58 @@
+'use client';
+
+import { Autocomplete as AutocompletePrimitive } from '@base-ui/react/autocomplete';
+import { cx } from 'class-variance-authority';
+import { type ReactNode } from 'react';
+import { getMatch } from '../menu/utils';
```
paanSinghCoder, line 6: Lets create a common util in `packages/raystack/utils` and move the shared util there. Menu and Command both will be using that shared util. Similar structure as `packages/raystack/hooks`.

paanSinghCoder (changes requested):

apps/www/src/content/docs/components/command/index.mdx
```diff
@@ -1,30 +1,162 @@
...
 
 ## Accessibility
 
-- Follows the [WAI-ARIA Combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)
-- Supports keyboard navigation with arrow keys and Enter to select
-- Uses `role="listbox"` for the command list and `role="option"` for items
+- Follows the [WAI-ARIA Combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/).
+- Input uses `role="combobox"`, list uses `role="listbox"`, items use `role="option"`.
+- Keyboard: `ArrowDown` / `ArrowUp` move the highlight, `Enter` activates the highlighted item, `Escape` closes the dialog.
+- Disabled items expose `aria-disabled="true"` and are skipped in keyboard navigation.
```
paanSinghCoder, line 161: There is no example of the `disabled` state in the docs. 

I think we can simply disable an item in the top most command example in the docs, no need to add a dedicated command for this as there is only one command component in the whole command docs page.

apps/www/src/components/playground/command-examples.tsx
```diff
@@ -1,29 +1,94 @@
...
+                  <Command.Item onClick={() => setOpen(false)}>
+                    Search Emoji
+                  </Command.Item>
+                  <Command.Item onClick={() => setOpen(false)}>
+                    Calculator
+                  </Command.Item>
+                </Command.Group>
+              </Command.List>
+            </Command>
+          </Command.Dialog.Content>
```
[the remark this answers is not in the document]
paanSinghCoder: @rohanchkrabrty Please check this.

apps/www/src/content/docs/components/command/index.mdx
rohanchkrabrty: `overlay` is a prop that accepts all backdrop props so it's exposing more than the className only.

Regarding the one off prop I agree, but `blur` is more like a feature even tho it changes the styling. Anyways we can decide on this later when we discuss on the one off props in general as a whole.

For now let's keep this to keep parity with `Dialog`

packages/raystack/components/command/command-misc.tsx
rohanchkrabrty: This component is to display keyboard shortcuts, where the input will always stay stable. `index` is fine for this component usage, let's not over-optimize the small things.

packages/raystack/components/command/command-item.tsx
rohanchkrabrty: Let's pick this up separately, as Command and Select also uses the same utils. There are other shared logic too which can be standardized at the same time

apps/www/src/content/docs/components/command/index.mdx
paanSinghCoder: Okay

paanSinghCoder (approved):

apps/www/src/components/playground/command-examples.tsx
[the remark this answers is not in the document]
rohanchkrabrty: Not relevant. It's an example in the internal playground page.

Issue #678, opened 2026-03-09
Remove !important declarations from component CSS


ravisuhag:
## Summary

Several component CSS modules use `!important` declarations to override specificity conflicts. This is a code smell indicating underlying CSS architecture issues and makes styles harder to override for consumers.

## Goal

Eliminate all `!important` declarations by fixing the underlying specificity issues through proper CSS cascade, selector specificity, or CSS Modules scoping.

## Affected Components

| Component | Issue | Details |
|-----------|-------|---------|
| Search (#632) | item 1 | Two `!important` flags (lines 17, 21) — specificity conflicts |
| FilterChip (#616) | item 1 | **11** `!important` declarations — heavy specificity issues |
| CodeBlock (#606) | item 1 | `!important` on `.token.comment` color in syntax theme |
| Calendar (#602) | item 3 | `.calendarPopover` uses `!important` for padding and border |

## Common Fixes

1. **Increase selector specificity** — use CSS Modules' `:local` scoping or add a parent class
2. **Reorder CSS** — ensure component styles load after base/reset styles
3. **Use `@layer`** — CSS cascade layers to control specificity without `!important`
4. **Flatten selectors** — reduce nesting that causes specificity wars

## Acceptance Criteria

- [ ] Zero `!important` declarations in component CSS modules
- [ ] Styles still render correctly without `!important`
- [ ] Consumer style overrides work without needing `!important`

rohanchkrabrty:
https://github.com/raystack/apsara/pull/751

PR #751, opened 2026-04-21
fix: remove unecessary important declarations

Shreyag02:
## Description

removed the unnecessary use of !important in css in reported components

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## How Has This Been Tested?

Dev testing

## Checklist:

## Related Issues

#678

### DIFF
--- a/packages/raystack/components/calendar/calendar.module.css
+++ b/packages/raystack/components/calendar/calendar.module.css
@@ -8,7 +8,7 @@
 }
 
 .captionLabel,
-.dropdowns>span,
+.dropdowns > span,
 .dropdown_trigger {
   font-weight: var(--rs-font-weight-medium);
   font-size: var(--rs-font-size-mini);
--- a/packages/raystack/components/calendar/calendar.module.css
+++ b/packages/raystack/components/calendar/calendar.module.css
@@ -170,13 +170,13 @@
   visibility: hidden;
 }
 
-.calendarPopover {
-  padding: var(--rs-space-3) !important;
+.calendarPopover.calendarPopover {
+  padding: var(--rs-space-3);
   /* Todo: var does not exist for 0.5px */
-  border: 0.5px solid var(--rs-color-border-base-primary) !important;
+  border: 0.5px solid var(--rs-color-border-base-primary);
   min-width: max-content;
   width: fit-content;
-  max-width: none !important;
+  max-width: none;
 }
 
 .dropdowns {
--- a/packages/raystack/components/calendar/calendar.module.css
+++ b/packages/raystack/components/calendar/calendar.module.css
@@ -170,6 +170,9 @@
   visibility: hidden;
 }
 
+/* Selector Doubling: chains the class with itself to raise specificity
+   from (0,1,0) to (0,2,0), beating react-day-picker's compound selectors
+   without resorting to !important or ID selectors. */
 .calendarPopover.calendarPopover {
   padding: var(--rs-space-3);
   /* Todo: var does not exist for 0.5px */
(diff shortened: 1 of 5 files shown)


packages/raystack/components/calendar/calendar.module.css
```diff
@@ -170,13 +170,13 @@
   visibility: hidden;
 }
 
-.calendarPopover {
-  padding: var(--rs-space-3) !important;
+.calendarPopover.calendarPopover {
```
rohanchkrabrty, line 176: can you check if this style block will work without this hack? If it works then let's remove it

packages/raystack/components/filter-chip/filter-chip.module.css
```diff
@@ -150,9 +150,15 @@ button.selectValue:hover {
   min-height: var(--rs-space-7);
 }
 
+.inputFieldWrapper [class*="input-field-module_inputWrapper"],
+.dateFieldWrapper [class*="input-field-module_inputWrapper"] {
+  border: none;
+  background: none;
+}
+
```
rohanchkrabrty: This styles are not needed. This is making the filter chip style behave in a diff way than intended. We will fix the input height bug in a separate PR.

packages/raystack/components/filter-chip/filter-chip.module.css
```diff
@@ -180,8 +186,13 @@ button.selectValue:hover {
...
-  padding-left: var(--rs-space-2) !important;
-  padding-right: var(--rs-space-2) !important;
+  padding-left: var(--rs-space-2);
+  padding-right: var(--rs-space-2);
+}
+
+.inputFieldWrapper input:focus,
+.dateFieldWrapper input:focus {
+  background: transparent;
 }
```
rohanchkrabrty: This styles are not needed. This is making the filter chip style behave in a diff way than intended. We will fix the input height bug in a separate PR.

rohanchkrabrty (changes requested):

packages/raystack/components/calendar/calendar.module.css
Shreyag02: The !important is needed because react-day-picker ships its own stylesheet where the popover/root is targeted with compound selectors (e.g. .rdp-root .rdp-…) at specificity (0,2,0). A plain .calendarPopover rule is only (0,1,0), so it loses the cascade and the padding doesn't apply — I confirmed it doesn't work without the bump.

Doubling to .calendarPopover.calendarPopover raises specificity to (0,2,0), tying the library's selectors. Since our stylesheet loads after react-day-picker's, source order wins the tie and our padding applies cleanly — no !important required.

This is also friendlier for consumers: with !important, any downstream override is forced into an !important arms race. With the specificity bump, consumers can override normally with any (0,2,1)+ selector (e.g. .myApp .calendarPopover) through the regular cascade.

packages/raystack/components/filter-chip/filter-chip.module.css
Shreyag02: removing and raising separate issue

packages/raystack/components/filter-chip/filter-chip.module.css
Shreyag02: removing and raising separate issue

packages/raystack/components/calendar/calendar.module.css
rohanchkrabrty: Yes, I understood why you did the double classname thing. I just wanted to confirm if it works with normal specificity or not. 

Since it doesn't lets keep this but can you add a comment over the double classname to something like "This is intended to increase css specificity"

rohanchkrabrty (approved):

PR #752, opened 2026-04-22
feat: RFC DataView component

rohanchkrabrty:
## Summary

- Adds RFC for DataView component
- Adds working beta for DataView component

### [**Preview Example**](https://apsara-ahvg28yvj-raystack.vercel.app/examples/dataview)


rohanchkrabrty:
1. Table & List View will use the same renderer `DataView.List` , 
  a. It will have support to toggle headers etc
  b. Consumers will handle the left, right division using a central cell in List View (can be rethinked if better implementation is there)
2. Renderer component will have option to override filterable, sort, display properties (check on dev implementation)
3. Each render will have a set of default display property
4. Check if tanstack table allows custom rendering composable

rohilsurana (approved):

PR #759, opened 2026-04-28
chore(plugin-vscode): migrate to OKLCH (color → culori)

paanSinghCoder:
## Summary

- Mirrored token data from HEX/rgba to `oklch()` to match the apsara CSS migration
- Swapped the `color` npm package for `culori` (the old one can't parse `oklch()`)

https://github.com/user-attachments/assets/6d2c10da-e2cb-449a-91c8-083aae0b1a79


rohilsurana:
https://www.npmjs.com/package/color-convert

rohanchkrabrty (approved):
Check if `color-convert` can be used directly to generate string, instead of computing the string. If we don't use `raw` function, the library handles the rounding part for us too.

paanSinghCoder:
@rohanchkrabrty Checked. No oklch stringifier. color-convert only stringifies rgb.hex ("FF0000"). Every oklch path returns a numeric tuple thus we still need the oklch(${L} ${C} ${H}) template.

PR good to go.

Issue #646, opened 2026-02-18
[ThemeProvider] Fix callback dependencies and add onThemeChange callback


ravisuhag:
## Summary

ThemeProvider is a custom theme management system with light/dark mode, localStorage persistence, and system preference detection. Has TypeScript safety issues, stale closure bugs, and obsolete CSS.

### 2. Fix missing useCallback dependencies
- `setTheme` callback (line 125) is missing `storageKey` in dependency array — can cause stale closure bugs
- `handleMediaQuery` callback (line 137) is missing `enableSystem` dependency

### 3. Fix invalid JavaScript string generation (line 278)
`val` construction uses `name + "|| ''"` which creates syntactically invalid JavaScript in the injected script.

### 5. Add `onThemeChange` callback
No event hook for applications that need to react to theme changes.

## Moved to Cross-Cutting Issues

The following items have been moved to cross-cutting issues that address them across all components:

- Replace hardcoded CSS values with design tokens (#672)
- TypeScript strictness (#675)
- Clean up deprecated code (#682)

## Files
- `packages/raystack/raystack/components/theme-provider/`

paanSinghCoder:
https://github.com/raystack/apsara/pull/760

PR #760, opened 2026-04-29
fix(theme-provider): callback deps, inline script, and onThemeChange

paanSinghCoder:
## Description

Addresses items 2, 3, and 5 from #646.

- **Callback deps** — `setTheme` now tracks `storageKey` (was `forcedTheme`, unused). `handleMediaQuery` now includes `enableSystem` and `applyTheme`. Both fix latent stale-closure bugs that surface only when those props change at runtime.
- **Inline-script fix** — `updateDOM` no longer emits `c.add(x[e]|| '')`. The literal path is gated with `if(${val})` so a missing `value`-map entry doesn't reach `classList.add('')` (throws) or `setAttribute(n, '')` (junk attr).
- **`onThemeChange` callback** — fires on actual `theme`/`resolvedTheme` changes, skips initial mount. Held via ref so consumer render churn doesn't over-fire and the callback never goes stale.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Improvement

## How Has This Been Tested?

Manual repro on `main` vs branch for each fix. Tests TODO — leaving as draft until `theme-provider.test.tsx` is extended.

## Related Issues

Refs #646 (items 2, 3, 5).

### DIFF
--- a/packages/raystack/components/theme-provider/theme.tsx
+++ b/packages/raystack/components/theme-provider/theme.tsx
@@ -167,30 +167,23 @@ const Theme = ({
     return () => window.removeEventListener('storage', handleStorage);
   }, [setTheme]);
 
-  // Whenever theme or forcedTheme changes, apply it
+  // Ref-held callback so consumer render churn doesn't drive effect cadence.
+  const onThemeChangeRef = useRef(onThemeChange);
+  onThemeChangeRef.current = onThemeChange;
+  const themeChangeMounted = useRef(false);
+
+  // Apply on theme/forcedTheme change, then notify (skipping initial mount).
   useEffect(() => {
     // @ts-ignore
     applyTheme(forcedTheme ?? theme);
-  }, [forcedTheme, theme]);
 
-  // Fire onThemeChange on actual changes, skipping the initial mount.
-  // Ref keeps the latest callback without re-firing when consumers pass an inline function.
-  const onThemeChangeRef = useRef(onThemeChange);
-  useEffect(() => {
-    onThemeChangeRef.current = onThemeChange;
-  });
-  const themeChangeMounted = useRef(false);
-  useEffect(() => {
-    if (!themeChangeMounted.current) {
-      themeChangeMounted.current = true;
-      return;
-    }
-    if (theme) {
+    if (themeChangeMounted.current && theme) {
       const resolved =
         theme === 'system' && resolvedTheme ? resolvedTheme : theme;
       onThemeChangeRef.current?.(theme, resolved);
     }
-  }, [theme, resolvedTheme]);
+    themeChangeMounted.current = true;
+  }, [forcedTheme, theme, resolvedTheme]);
 
   const providerValue = useMemo(
     () => ({
--- a/packages/raystack/components/theme-provider/theme.tsx
+++ b/packages/raystack/components/theme-provider/theme.tsx
@@ -54,8 +54,8 @@ const Theme = ({
   const [theme, setThemeState] = useState(() =>
     getTheme(storageKey, defaultTheme)
   );
-  const [resolvedTheme, setResolvedTheme] = useState(() =>
-    getTheme(storageKey)
+  const [resolvedTheme, setResolvedTheme] = useState<string | undefined>(
+    undefined
   );
   const attrs = !value ? themes : Object.values(value);
 
--- a/packages/raystack/components/theme-provider/theme.tsx
+++ b/packages/raystack/components/theme-provider/theme.tsx
@@ -178,7 +178,8 @@ const Theme = ({
     applyTheme(forcedTheme ?? theme);
 
     if (!theme) return;
-    const resolved = theme === 'system' ? resolvedTheme : theme;
+    const resolved =
+      forcedTheme ?? (theme === 'system' ? resolvedTheme : theme);
     if (!resolved) return;
 
     if (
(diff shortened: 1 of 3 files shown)


packages/raystack/components/theme-provider/theme.tsx
```diff
@@ -293,13 +314,17 @@ const ThemeScript = memo(
       }
 
       if (attribute === 'class') {
-        if (literal || resolvedName) {
+        if (literal) {
+          text += `if(${val})c.add(${val})`;
```
rohanchkrabrty, line 322: Is the `if` block needed in the generated script string?

packages/raystack/components/theme-provider/theme.tsx
```diff
@@ -171,6 +173,25 @@ const Theme = ({
...
+      themeChangeMounted.current = true;
+      return;
+    }
+    if (theme) {
+      const resolved =
+        theme === 'system' && resolvedTheme ? resolvedTheme : theme;
+      onThemeChangeRef.current?.(theme, resolved);
+    }
+  }, [theme, resolvedTheme]);
+
```
rohanchkrabrty: we can call the `onThemeChange` in `applyTheme` or in `applyTheme` useEffect.
Right now there are too many useEffects

packages/raystack/components/theme-provider/theme.tsx
```diff
@@ -171,6 +173,25 @@ const Theme = ({
     applyTheme(forcedTheme ?? theme);
   }, [forcedTheme, theme]);
 
+  // Fire onThemeChange on actual changes, skipping the initial mount.
+  // Ref keeps the latest callback without re-firing when consumers pass an inline function.
+  const onThemeChangeRef = useRef(onThemeChange);
```
rohanchkrabrty, line 171: Is this needed?, on each render the ref is getting updated

rohanchkrabrty (changes requested):

packages/raystack/components/theme-provider/theme.tsx
paanSinghCoder: Yes, it's needed when consumers pass a value map, `x[e]` can be undefined if the user's stored theme isn't a key in the map, and without the guard `c.add(undefined)` adds a literal "undefined" class. Without value it's redundant since `else if(e)` already gates `e` truthy.

packages/raystack/components/theme-provider/theme.tsx
paanSinghCoder: Good catch. Added the notify into the apply-theme useEffect and added `resolvedTheme` to its deps.

packages/raystack/components/theme-provider/theme.tsx
paanSinghCoder: The ref is needed. without it `onThemeChange` will have to go in the effect's deps. Consumers passing inline callbacks will then re-run the effect every render and fire the callback repeatedly with the same theme.

The useEffect wrapper around the ref update wasn't needed though. replaced with a render-time `onThemeChangeRef.current = onThemeChange`;

packages/raystack/components/theme-provider/theme.tsx
rohanchkrabrty: Understood

packages/raystack/components/theme-provider/theme.tsx
```diff
@@ -165,11 +167,23 @@ const Theme = ({
...
+  onThemeChangeRef.current = onThemeChange;
+  const themeChangeMounted = useRef(false);
+
+  // Apply on theme/forcedTheme change, then notify (skipping initial mount).
   useEffect(() => {
     // @ts-ignore
     applyTheme(forcedTheme ?? theme);
-  }, [forcedTheme, theme]);
+
+    if (themeChangeMounted.current && theme) {
```
rohanchkrabrty: Since resolvedTheme creates a second render, this effect is run twice and the onThemeChange will be called in initial mount itself.

packages/raystack/components/theme-provider/theme.tsx
```diff
@@ -165,11 +167,23 @@ const Theme = ({
...
+      const resolved =
+        theme === 'system' && resolvedTheme ? resolvedTheme : theme;
+      onThemeChangeRef.current?.(theme, resolved);
+    }
+    themeChangeMounted.current = true;
```
rohanchkrabrty: Can we skip this assignment once done?

packages/raystack/components/theme-provider/theme.tsx
```diff
@@ -165,11 +167,23 @@ const Theme = ({
...
+  // Apply on theme/forcedTheme change, then notify (skipping initial mount).
   useEffect(() => {
     // @ts-ignore
     applyTheme(forcedTheme ?? theme);
-  }, [forcedTheme, theme]);
+
+    if (themeChangeMounted.current && theme) {
+      const resolved =
+        theme === 'system' && resolvedTheme ? resolvedTheme : theme;
+      onThemeChangeRef.current?.(theme, resolved);
```
rohanchkrabrty, line 194: Claude flagged that setting forcedTheme wont trigger onThemeChange. Please check once

rohanchkrabrty (changes requested):
Also let's update the docs in `theme/overview` with the new props

packages/raystack/components/theme-provider/theme.tsx
paanSinghCoder: Fixed

packages/raystack/components/theme-provider/theme.tsx
paanSinghCoder: Done

packages/raystack/components/theme-provider/theme.tsx
paanSinghCoder: Fixed

rohanchkrabrty (approved):

PR #764, opened 2026-04-29
feat: Toast support for leadingIcon and update default styles

rohanchkrabrty:
## Summary
- Add `leadingIcon` prop to `toastManager.add/update` and `Toast.createToastManager`, lifting it onto Base UI's typed `data` slot via a wrapper.
- Map success/error/warning/info/loading toast types to default Radix icons (with apsara `Spinner` for loading); explicit `leadingIcon` still wins.
- Restructure toast layout to match Figma: header row (icon + title + actions) and a separate description row indented to align under the title; description color updated to `foreground-base-primary`.
- Title-only toasts render with the description text style so they don't look outsized; only the icon is colored by type — the surface stays neutral.
- Replace the static demo at the top of the toast docs with an interactive playground (title, description, type, actionButton toggle) and add tests for the new behaviors.

### DIFF
--- a/apps/www/src/content/docs/components/toast/props.ts
+++ b/apps/www/src/content/docs/components/toast/props.ts
@@ -70,6 +70,12 @@ export interface ToastManagerAddOptions {
   actionProps?: React.ComponentPropsWithoutRef<'button'>;
 
   /**
+   * Icon rendered before the toast title. Inherits color from the toast type
+   * (e.g. green for `type: "success"`).
+   */
+  leadingIcon?: React.ReactNode;
+
+  /**
    * Custom data to attach to the toast.
    */
   data?: Record<string, unknown>;
--- a/apps/www/src/content/docs/components/toast/props.ts
+++ b/apps/www/src/content/docs/components/toast/props.ts
@@ -71,17 +71,137 @@ export interface ToastManagerAddOptions {
 
   /**
    * Icon rendered before the toast title. Inherits color from the toast type
-   * (e.g. green for `type: "success"`).
+   * (e.g. green for `type: "success"`). Omit to use the default icon for the
+   * toast type, pass any React node to override it, or pass `null` to render
+   * no icon at all.
    */
   leadingIcon?: React.ReactNode;
-
-  /**
-   * Custom data to attach to the toast.
-   */
-  data?: Record<string, unknown>;
 
   /**
    * Optional custom ID for the toast. Auto-generated if not provided.
    */
   id?: string;
+}
+
+/**
+ * Options for `update(id, options)`. Same shape as `ToastManagerAddOptions`,
+ * but every field is optional and the toast `id` moves to the first argument.
+ */
+export interface ToastManagerUpdateOptions {
+  title?: React.ReactNode;
+  description?: React.ReactNode;
+  type?: 'success' | 'error' | 'info' | 'warning' | 'loading';
+  timeout?: number;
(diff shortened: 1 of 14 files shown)


packages/raystack/components/toast/toast-root.tsx
```diff
@@ -39,6 +57,12 @@ export function ToastRoot({
 }: ToastRootProps) {
   const swipeDirection = getSwipeDirection(position);
   const hasDescription = !!toast.description;
+  const explicitLeadingIcon = (toast.data as ToastData | undefined)
+    ?.leadingIcon;
+  const leadingIcon =
+    explicitLeadingIcon ??
+    (toast.type ? TOAST_ICONS[toast.type] : null) ??
+    TOAST_ICONS.default;
```
paanSinghCoder: We always render the `InfoCircledIcon` as default if `leadingIcon` is not passed. What if the user doesn't want to render any leading icon?

This is also a breaking change as consumers are not expecting a `leadingIcon` by default.

apps/www/src/content/docs/components/toast/props.ts
paanSinghCoder: Here data is mentioned as custom data that can be attached to the toast but in we have omitted data `Omit<BaseAddOptions, 'data'> & {`

Image


Tried it in `examples/page.tsx`. This is also flagged in the IDE.

packages/raystack/components/toast/toast.tsx
```diff
@@ -1,10 +1,11 @@
 import { Toast as ToastPrimitive } from '@base-ui/react';
+import { createToastManager } from './toast-manager';
 import { ToastProvider } from './toast-provider';
 import { ToastRoot } from './toast-root';
 
 export const Toast = Object.assign(ToastRoot, {
   Provider: ToastProvider,
-  createToastManager: ToastPrimitive.createToastManager,
+  createToastManager,
   useToastManager: ToastPrimitive.useToastManager
```
paanSinghCoder: `useToastManager` missing in the docs.

packages/raystack/components/toast/toast-manager.ts
```diff
@@ -1,5 +1,98 @@
...
+  leadingIcon?: ReactNode;
+}
+
+type BaseManager = ReturnType<
+  typeof ToastPrimitive.createToastManager<ToastData>
+>;
+type BaseAddOptions = Parameters<BaseManager['add']>[0];
+type BaseUpdateOptions = Parameters<BaseManager['update']>[1];
+
+export type ToastAddOptions = Omit<BaseAddOptions, 'data'> & {
```
paanSinghCoder: Breaking change to omit `data`. We can add it in migration guide and also as suggested above, we can remove this from docs.

packages/raystack/components/toast/toast-root.tsx
```diff
@@ -49,34 +73,43 @@ export function ToastRoot({
       {...props}
     >
       <ToastPrimitive.Content className={styles.content}>
-        <Flex direction='column' gap={1} className={styles.textContainer}>
-          {toast.title && (
-            <ToastPrimitive.Title
-              className={hasDescription ? styles.title : styles.description}
+        <div className={styles.header}>
+          <div className={styles.headerLeft}>
```
paanSinghCoder: Flex can be used on line 76 and 77. Not a compulsion here as its not adding too much value but good to have.

paanSinghCoder (approved):

PR #766, opened 2026-04-29
refactor: refactor Label with useRender; reuse in Field.Label

rohanchkrabrty:
## Summary
- Rebuild `Label` as a polymorphic, `useRender`-based component. Drops the old `size` variants and asterisk-style `required`/`requiredIndicator`; new props are `required` (renders an `(optional)` indicator when `false`) and `optionalText` (defaults to `'(optional)'`). CSS is typography-only with no margins/padding — pointer cursor is wired via `.label[for]` so layout is the consumer's responsibility (`Flex`).
- `Field.Label` delegates to `Label` through the Base UI `render` prop. `FieldLabelProps = FieldPrimitive.Label.Props & LabelProps`, so future Label props flow through automatically; the `(optional)` indicator now lives in Label.
- Aligns Field spacing with Figma (node 4908:11107): `.field { gap: var(--rs-space-2) }` (was `--rs-space-1`); strips `padding-bottom` from the label and `padding-top` from `.description`/`.error`; removes the now-redundant `.label`/`.optional` rules from `field.module.css`.
- Migrates raw `<label>` callsites to `Label` (theme-customiser radio rows); updates Radio/Checkbox/Label playground examples and docs (`index.mdx`, `props.ts`, `demo.ts`) for the new API.
- Bumps `.scrub-area-label[for]` specificity in NumberField so `cursor: ew-resize` wins over Label's `[for]` pointer rule. Label tests rewritten (11 cases); existing Field/NumberField tests still pass.


rohilsurana (approved):
LGTM — clean consolidation of Label into Field.Label with good test coverage.

A few minor suggestions for consideration (non-blocking):

- Commit prefix should probably be `feat!:` instead of `chore:` since removing the `Label` export is a breaking change for consumers.
- The fallback `<label>` path could strip Base UI-specific props before spreading to avoid potential DOM warnings.
- Consider a note in Field docs about `Field.Label orientation="horizontal"` as the replacement pattern for standalone Label usage with Radio/Checkbox.

Shreyag02 (approved):

rsbh (approved):

PR #768, opened 2026-04-30
feat(theme): scoped theming with ThemeScope component

paanSinghCoder:
## Summary

https://github.com/user-attachments/assets/2dc4af63-f10f-4293-9f41-f73ffbc2b97a

Scoped theming for Apsara — a nested `<Theme>` creates an isolated themed region inside a page. Example: `/examples/theme-overrides`.

## What's new

- **Inherits from the parent by default.** A nested `<Theme>` with no props is invisible — children render exactly as they would without it. You only pay for what you override.

- **Pick which axes to override.** Pass any combination of theme (light/dark), accent color, gray, or style variant. Anything you don't pass keeps inheriting from the parent.

- **Remembers user choices when you give it a `storageKey`.** Toggle a region's theme and it sticks — survives reloads, syncs across tabs automatically. Clear it to fall back to inheriting again.

- **`Theme` is the canonical name.** `ThemeProvider` keeps working as a deprecated alias; no migration required, but new code should use `Theme`.

- **Components inside a region pick up the local theme.** Buttons, callouts, shadows, native scrollbars, and the smooth theme-switch transition all follow the nearest scope — no extra plumbing in component code.

## Test plan

- [ ] `pnpm test` in `packages/raystack` — 38 theme-provider tests pass.
- [ ] Visit `/examples/theme-overrides` — every mode has inline test steps and expected behavior.
- [ ] Bare `<div data-theme="dark">` inside a light page — native UI (checkboxes, scrollbars), shadows, and accent tokens all follow the scope.
- [ ] Toggle a persistent scope panel, refresh — the choice survives. Open in two tabs — toggles propagate. Click Clear — scope re-inherits.


packages/raystack/styles/primitives/accent.css
```diff
@@ -1,11 +1,12 @@
...
- * It is used for generating the final colors file. 
+/*
+ * Note: Do not use this file directly.
+ * It is used for generating the final colors file.
  * All variables must be used from colors.css file only.
  */
 
 /* [data-accent-color='indigo']  { */
-:root {
+:root,
```
rohanchkrabrty: Let's remove the default `:root` selector

packages/raystack/styles/primitives/appearance.css
```diff
@@ -4,26 +4,38 @@
...
-    background-color 0.4s ease,
-    color 0.4s ease;
+/* Smooth theme switch transition — applies regardless of which theme is active */
+[data-theme="light"],
+[data-theme="dark"] {
+    transition: background-color 0.4s ease, color 0.4s ease;
+}
+
+/* Native UI (form controls, scrollbars, text selection) follows the scoped theme */
+:root,
```
rohanchkrabrty, line 16: Let's remove the default `:root` selector

packages/raystack/styles/primitives/appearance.css
```diff
@@ -4,26 +4,38 @@
...
-  --rs-neutral-9: var(--rs-gray-9);
-  --rs-neutral-10: var(--rs-gray-10);
-  --rs-neutral-11: var(--rs-gray-11);
-  --rs-neutral-12: var(--rs-gray-12);
+[data-theme="dark"] {
+    color-scheme: dark;
+}
+
+/* Light Theme Colors */
+:root,
```
rohanchkrabrty: Let's remove the default `:root` selector

packages/raystack/styles/primitives/gray.css
```diff
@@ -5,7 +5,9 @@
  */
 
 
-[data-gray-color='gray'] {
+:root,
```
rohanchkrabrty: Let's remove the default `:root` selector

packages/raystack/styles/primitives/gray.css
```diff
@@ -5,7 +5,9 @@
...
+[data-gray-color='gray'],
+[data-theme='light'] {
```
rohanchkrabrty: Why are we using this `[data-theme='light']` here?

packages/raystack/styles/primitives/gray.css
```diff
@@ -20,7 +22,8 @@
     --rs-gray-12: oklch(0.2435 0 0);
 }
 
-[data-gray-color='gray'][data-theme='dark'] {
+[data-gray-color='gray'][data-theme='dark'],
+[data-theme='dark'] {
```
rohanchkrabrty: Why are we using this `[data-theme='dark']` here?

packages/raystack/styles/colors.css
```diff
@@ -1,9 +1,11 @@
 /*
  * Note: Use color variables from this file only.
- * Avoid using color variables from other files in the v1/styles folder directly.
+ * Avoid using color variables from other files in the styles folder directly.
  */
 
-:root {
+:root,
```
rohanchkrabrty: Let's remove the default `:root` selector

packages/raystack/components/theme-provider/theme.tsx
rohanchkrabrty, line 1: 1. There is no point of branching out Stateless Scope. The main difference is storageKey which can be handled in a single component. Both the scope component can be merged and reduce logic dupe
2. The expectation was that nested theme provider would only apply the overriden styles, but inherit rest of the stuff. But in provided example, if `accentColor` is overriden, it's theme is not getting inherited from the root.
3. The choice of `useTheme` not working in Stateless Scope is misleading. Local storage persistance shouldn't affect hook usage. Right now the `useTheme` is ambiguous, from a call site a user cannot understand what will happen. It should ideally always update the nearest `ThemeContext`.
    a.  If updating parent scope needs to be feature, then we can introduce something like `useTheme({storageKey})` and use that pattern to target outside ThemeContexts

packages/raystack/components/theme-provider/theme.tsx
```diff
@@ -11,30 +10,194 @@ import {
   useRef,
   useState
 } from 'react';
-
 import type { ThemeProviderProps, UseThemeProps } from './types';
+import { COLOR_SCHEMES } from './types';
 
-const colorSchemes = ['light', 'dark'];
+const colorSchemes: string[] = [...COLOR_SCHEMES];
```
rohanchkrabrty: Can't we use the `COLOR_SCHEMES` directly?

apps/www/src/content/docs/theme/overview/index.mdx
rohanchkrabrty, line 1: All examples use `ThemeProvider` alias even tho it's deprecated

rohanchkrabrty (dismissed):

packages/raystack/styles/primitives/accent.css
paanSinghCoder: It was there as there might be user who doesn't want to use our `Theme`, in this case the `:root` variables applies by default. But as we have mentioned in docs to use Apsara's theme, we should remove it as not needed.

Done.

packages/raystack/components/theme-provider/theme.tsx
paanSinghCoder: Directly using it will throw error. `TypeScript error. Array.prototype.includes` is strictly typed on tuples, the search argument must match an element type. Removed the unnecessary spread though.

packages/raystack/styles/primitives/appearance.css
paanSinghCoder: Done

packages/raystack/styles/primitives/appearance.css
paanSinghCoder: Done

packages/raystack/styles/primitives/gray.css
paanSinghCoder: Done

packages/raystack/styles/primitives/gray.css
paanSinghCoder: Without it, a scoped wrapper that has only data-theme wouldn't get a local `--rs-gray-*` definition and would inherit the parent's gray scale instead.

packages/raystack/styles/colors.css
paanSinghCoder: Done.

apps/www/src/content/docs/theme/overview/index.mdx
paanSinghCoder: Done.

packages/raystack/styles/primitives/gray.css
paanSinghCoder: Same reason as above.

packages/raystack/components/theme-provider/theme.tsx
paanSinghCoder: Done.

paanSinghCoder: done

rsbh (approved):

Issue #674, opened 2026-03-09
Forward HTML attributes (spread rest props)


ravisuhag:
## Summary

Many components selectively pick props instead of extending HTML attributes and spreading the rest. This prevents consumers from passing standard HTML attributes like `className`, `style`, `data-*`, `aria-*`, `onClick`, etc.

## Goal

All components should extend the appropriate `HTMLAttributes` interface and spread `...rest` props onto their root DOM element.

## Pattern

```tsx
// Before
function Component({ label, size }: ComponentProps) {
  return <div>{label}</div>;
}

// After
function Component({ label, size, ...rest }: ComponentProps) {
  return <div {...rest}>{label}</div>;
}
```

Where `ComponentProps` extends `React.HTMLAttributes<HTMLDivElement>` (or the appropriate element type).

## Affected Components

| Component | Issue | Details |
|-----------|-------|---------|
| Indicator | #622 | Should spread remaining props onto root |
| Image | #621 | Missing native `<img>` attributes (loading, decoding, fetchPriority) |
| IconButton | #620 | Should spread onto underlying button |
| Headline | #619 | Should spread onto rendered heading element |
| FilterChip | #616 | Should spread onto root element |
| EmptyState | #615 | Should spread onto root element |
| CopyButton | #611 | Should spread onto underlying IconButton |
| Chip | #605 | Props selectively picked instead of spread |
| Badge | #598 | Can't pass `onClick`, `style`, `data-*`, `role`, etc. |
| Amount | #595 | Can't pass `className`, `style`, `data-*`, `aria-*`, `onClick` |

## Acceptance Criteria

- [ ] All components extend appropriate `HTMLAttributes` interface
- [ ] All components spread `...rest` props onto root DOM element
- [ ] TypeScript autocomplete shows available HTML attributes
- [ ] No breaking changes to existing prop APIs

paanSinghCoder:
https://github.com/raystack/apsara/pull/769

Issue #622, opened 2026-02-18
[Indicator] Add animation support


ravisuhag:
## Summary

Indicator is a custom status indicator component. Has inconsistent token usage.

### 3. Add animation support
Consider adding a `pulse` or `animated` prop for active/live indicators.

## Moved to Cross-Cutting Issues

The following items have been moved to cross-cutting issues that address them across all components:

- Replace hardcoded CSS values with design tokens (#672)
- Forward HTML attributes (#674)

## Files
- `packages/raystack/raystack/components/indicator/`

Issue #621, opened 2026-02-18
[Image] Add forwardRef and remove redundant accessibility attributes


ravisuhag:
## Summary

Image is a custom component wrapping the native `<img>` element. Has redundant attributes and missing ref forwarding.

### 1. Add `forwardRef` support
Component doesn't forward refs, preventing parent components from accessing the underlying `<img>` element.

### 2. Remove redundant accessibility attributes
Some accessibility attributes duplicate native HTML behavior and can be removed.

### 3. Forward HTML img attributes
Should spread remaining props onto the `<img>` element to support all native attributes (loading, decoding, fetchPriority, etc.).

### 4. Add error/loading state handling
No built-in handling for image load errors or loading states. Consider adding `onError` fallback support.

## Files
- `packages/raystack/raystack/components/image/`

ravisuhag:
Verified against PR #856 (commit `dd140b54`).

`image.tsx` now handles the load-error and loading cases this issue asks for:

- **One-shot fallback** — `onError` assigns `fallback` once. If the fallback itself fails to load, it is not re-assigned again (no infinite error loop). A test covers this.
- **No fallback case** — when there is no fallback, or the fallback also fails, the image falls back to the normal alt/broken rendering instead of staying invisible.
- **Load state** — the image tracks `loading`/`loaded` and fades in once decoded. Already-decoded (cached or SSR-painted) images skip the fade, so there is no flash.
- `onLoad` is now forwarded too.

Proposing to close this once #856 merges.

Issue #620, opened 2026-02-18
[IconButton] Component improvements


ravisuhag:
## Summary

IconButton is a custom button component for icon-only actions. Has redundant accessibility attributes and minor improvements.

## Moved to Cross-Cutting Issues

The following items have been moved to cross-cutting issues that address them across all components:

- Accessibility baseline (#673)
- Forward HTML attributes (#674)

## Files
- `packages/raystack/raystack/components/icon-button/`

PR #769, opened 2026-05-04
fix(chip,filter-chip): forward HTML attributes onto root element

paanSinghCoder:
## Summary

Closes the remaining gap from #674 — covers **Chip** (#605) and **FilterChip** (#616). The other components listed in #674 already extend the appropriate `HTMLAttributes` type and spread `...rest` (verified against `main` at `a7e6705`).

- **Chip** had a real runtime gap: it selectively picked `onClick`, `role`, `data-state`, `disabled`, `ariaLabel` and dropped everything else, so consumers couldn't pass `style`, `id`, `data-*`, `aria-*`, mouse events, etc.
- **FilterChip** already spread `...props` onto the root `<Flex>` at runtime, but `FilterChipProps` didn't extend any HTMLAttributes type — so TypeScript rejected those props at the call site.

Both now extend `ComponentProps<'span'>` / `ComponentProps<'div'>` and the Chip implementation spreads the rest props onto the root element. Existing prop API is preserved:

- `ariaLabel` (legacy camelCase) still wins over the native `aria-label`
- `disabled` still gates `onClick` and sets `data-disabled`
- `role` still defaults to `'status'` on Chip
- `data-state` is still passed through (now via the spread instead of an explicit prop)

## Components NOT changed (already compliant)

| Component | Status |
|---|---|
| Indicator (#622) | already extends `ComponentProps<'div'>` and spreads |
| Image (#621) | already extends `ImgHTMLAttributes` |
| IconButton (#620) | already extends `ComponentProps<'button'>` and spreads |
| Headline (#619) | uses Base UI `useRender.ComponentProps<'h2'>` + `mergeProps` |
| EmptyState (#615) | already extends `ComponentProps<typeof Flex>` |
| CopyButton (#611) | already extends `IconButtonProps` |
| Badge (#598) | already extends `ComponentProps<'span'>` |
| Amount (#595) | already extends `ComponentProps<'span'>` |

If maintainers agree, #674 can be closed once this lands.

## Test plan

- [x] `pnpm test` — full suite passes (1684 tests; +4 new covering arbitrary attribute forwarding on Chip and FilterChip)
- [x] `pnpm build` — typecheck clean
- [x] Existing `ariaLabel`, `role`, `disabled`, dismiss, and click behavior covered by existing Chip tests still pass
- [x] Verify in Storybook/docs that consumers can now pass `style`, `data-*`, `id`, `onMouseEnter`, etc.

### DIFF
--- a/packages/raystack/components/chip/chip.tsx
+++ b/packages/raystack/components/chip/chip.tsx
@@ -34,7 +34,6 @@ type ChipProps = ComponentProps<'span'> &
     isDismissible?: boolean;
     children: ReactNode;
     onDismiss?: () => void;
-    ariaLabel?: string;
     disabled?: boolean;
   };
 
@@ -50,9 +49,8 @@ export const Chip = ({
   onDismiss,
   onClick,
   role = 'status',
-  ariaLabel,
   disabled,
-  'aria-label': ariaLabelAttr,
+  'aria-label': ariaLabel,
   ...props
 }: ChipProps) => {
   const handleDismiss = (e: React.MouseEvent) => {
@@ -66,9 +64,7 @@ export const Chip = ({
       className={chip({ variant, size, color, className })}
       role={role}
       aria-label={
-        ariaLabel ??
-        ariaLabelAttr ??
-        (typeof children === 'string' ? children : undefined)
+        ariaLabel ?? (typeof children === 'string' ? children : undefined)
       }
       onClick={disabled ? undefined : onClick}
       data-disabled={disabled || undefined}
--- a/packages/raystack/components/chip/chip.tsx
+++ b/packages/raystack/components/chip/chip.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { cva, type VariantProps } from 'class-variance-authority';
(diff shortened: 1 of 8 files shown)


paanSinghCoder:
Closing for now — keeping changes local for further testing.

packages/raystack/components/chip/chip.tsx
```diff
@@ -27,19 +27,16 @@ const chip = cva(styles.chip, {
...
-  'data-state'?: string;
-};
+type ChipProps = ComponentProps<'span'> &
+  VariantProps<typeof chip> & {
+    trailingIcon?: ReactNode;
+    leadingIcon?: ReactNode;
+    isDismissible?: boolean;
+    children: ReactNode;
+    onDismiss?: () => void;
+    ariaLabel?: string;
```
rsbh: we are supporting both `ariaLabel` and `aria-label`

paanSinghCoder: Done. Removed ariaLabel as aria-label is the standard we use.

rsbh (approved):

Issue #675, opened 2026-03-09
TypeScript strictness: eliminate any, @ts-ignore, and export prop types


ravisuhag:
## Summary

Several components have TypeScript safety issues: `any` types, `@ts-ignore` suppressions, missing type exports, unsafe type casting, and outdated type patterns. These reduce type safety for both maintainers and consumers.

## Goal

Achieve strict TypeScript compliance across all components with zero `any` types, zero `@ts-ignore` comments, and full prop type exports.

## Affected Components

| Component | Issue | Details |
|-----------|-------|---------|
| Tooltip (#648) | item 2 | Unsafe index operation in `getTransformForPlacement` — `align` could be undefined |
| Toast (#647) | items 1, 4 | `toast()` uses wrong type (`ToasterProps` vs `ToastOptions`); empty interface extension |
| ThemeProvider (#646) | item 1 | Two `@ts-ignore` comments without explanation |
| Sidebar (#637) | item 4 | `as?: ReactElement` too permissive — should be `React.ElementType` |
| SidePanel (#636) | item 2 | `SidePanelProps` doesn't extend `HTMLAttributes<HTMLAsideElement>` |
| Radio (#630) | item 3 | `RadioProps` and `RadioGroupProps` not exported |
| List (#626) | item 2 | `align` prop declared in docs but missing from TypeScript interface |
| DataTable (#612) | items 1, 3 | `as unknown as TData[]` unsafe casting; column defs not memoized |
| Command (#609) | item 1 | `export const Command: any` — loses all type safety |
| Breadcrumb (#600) | item 21 | `BreadcrumbProps` extends `HTMLDivElement` but renders `<nav>` |
| Amount (#595) | item 7 | `@ts-ignore` on `format()` call |
| Accordion (#592) | item 6 | Deprecated `ElementRef` should be `ComponentRef` |
| Checkbox (#604) | item 16 | Deprecated `ElementRef` should be `ComponentRef` |

### 1. Replace `any` with proper types
```tsx
// Before
export const Command: any = Object.assign(...)
// After
export const Command: CommandType = Object.assign(...)
```

### 2. Remove `@ts-ignore` with proper type handling
```tsx
// Before
// @ts-ignore
format(finalBaseValue)
// After
format(Number(finalBaseValue))
```

### 3. Export prop types for consumer use
```tsx
// Add to public API
export type { RadioProps, RadioGroupProps } from './radio';
```

### 4. Migrate `ElementRef` → `ComponentRef`
```tsx
[...]

PR #770, opened 2026-05-04
typescript strictness cleanup

Shreyag02:
## Description

Cleanup pass that strengthens type-safety across the design system and the docs site without changing runtime behavior in production code paths.                                                                                                                                         
                                                                                                                                                 
  Three layers of work:                                                                                                                          
                                                                                                                                                 
  1. Removed unsafe suppressions: 7 @ts-ignore and 6 @ts-expect-error lines either deleted (where the underlying issue was fixable) or replaced  
  with proper APIs / @ts-expect-error (where the upstream type gap is real).
  2. Replaced any with precise types: ~110 any usages removed (114 → ~10) across raystack components, data-table internals, app helpers, demo    
  files, and tests. New types added where needed: FilterPrimitive union, MdxJsxFlowElementLike interface, ComponentPropsType reuse.              
                                          
  Notable correctness improvements (latent bug fixes, not regressions):                                                                          
  - theme-provider: el.style.colorScheme = null (which browsers stringified to "null") → ?? '' correctly clears the property
  - mdx-components.tsx: heading mappers now use render={<hN />} (BaseUI's actual render-prop API), 
  so `<h1>`–`<h6>` render as the correct semantic  tags. Previously rendered as `<h2>`/ `<span>` with a stray as="hN" HTML attribute.                            
  - examples/table/page.tsx: added ?? [] null guard so the status filter's .includes() no longer throws on undefined filter values.              
                                                                                                                                   
  Tooling cleanup:                                                                                                                               
  - Replaced the local apps/www/src/types/prettier.d.ts shim with @types/prettier@^2 as a devDependency.      

## Type of Change

- [ ] Documentation update

## How Has This Been Tested?

  - tsc --noEmit run before and after on both packages/raystack and apps/www. Error counts unchanged from main baseline (raystack 13/13, apps/www
   56/54). The 2-error drop in apps/www comes from the mdx-components.tsx render prop fix clearing 6 @ts-expect-error suppressions.
  - biome check on all changed files. No new lint warnings introduced; remaining 4 warnings are pre-existing (React unused import in filters.tsx,
   deliberately-kept any in filter-chip.tsx).                                                                                                    
  - Spot-checked all 24 newly-introduced as/as unknown as casts: each is either inside a runtime-narrowed branch (e.g., case FilterType.X) or in
  test code asserting deliberately-invalid input.                                                                                                
  - Manually traced consumers of DataTableFilterValues.value and FilterChip.value to confirm cascade casts are correct at every use site (2 cascades in filters.tsx and examples/table/page.tsx were caught and addressed).    

## Checklist:

  N/A — no UI changes. The mdx-components.tsx render prop fix changes the rendered HTML tag from `<h2>`/`<span>` to `<h1>`–`<h6>` for semantic           
  correctness; visual styling stays identical because it's driven by the size prop, not the tag name.

## Related Issues

#675


apps/www/src/components/mdx/mdx-components.tsx
```diff
@@ -66,6 +66,7 @@ const mdxComponents = {
   img: Image,
   h1: (props: HTMLAttributes<HTMLHeadingElement>) => (
     <Headline
+      //@ts-expect-error - Headline component type doesn't support h1 element
       as='h1'
```
rohanchkrabrty: we can use `render` prop here as `as` support was removed. TS won't show error then

apps/www/src/types/prettier.d.ts
rohanchkrabrty, line 1: Add `@types/prettier@^2` as dev dependency instead of local shim

rohanchkrabrty (changes requested):

paanSinghCoder (approved):

apps/www/src/components/mdx/mdx-components.tsx
Shreyag02: resolved

packages/raystack/CHANGELOG.md
rohanchkrabrty, line 1: let's drop this changes. We currently don't maintain CHANGELOG

apps/www/src/types/prettier.d.ts
Shreyag02: updated

packages/raystack/components/command/command-root.tsx
rohanchkrabrty, line 1: Let's drop this changes. Base UI Autocomplete uses `any` so let's keep parity with that

apps/www/src/content/docs/components/form/props.ts
rohanchkrabrty, line 1: Let's drop this change to keep parity with Base UI

apps/www/src/content/docs/components/tabs/props.ts
rohanchkrabrty, line 1: Let's keep it any to keep parity with Base UI.

Base UI components accept any or Generic, so let';s not type it to string

apps/www/src/content/docs/components/radio/props.ts
rohanchkrabrty, line 1: Let's keep it any to keep parity with Base UI.

Base UI components accept any or Generic, so let';s not type it to string

packages/raystack/components/filter-chip/filter-chip.tsx
rohanchkrabrty: the onValueChange also needs to be updated. make a reusable type and use in both Value and onValueChange

rohanchkrabrty (changes requested):

rohanchkrabrty (approved):

PR #774, opened 2026-05-05
fix(data-table): make virtualized loader rows scroll with content

paanSinghCoder:
Loader skeleton rows in `DataTable.VirtualizedContent` were pinned to the bottom via `position: sticky`. They should scroll with the list and render at the end of the loaded rows.

## Test plan
- [ ] Open the virtualized DataTable example, trigger load-more, and confirm skeleton rows appear at the end of the list and scroll with the content (no longer pinned to the viewport bottom).


paanSinghCoder:
Replaced by #776 (same diff, renamed branch).

PR #777, opened 2026-05-07
fix(data-table): correct toolbar horizontal padding

rohanchkrabrty:
## Summary
- Fix data-table toolbar horizontal padding to match the Figma spec.
- Replace the single-value `padding: var(--rs-space-3)` on `.toolbar` with the 4-value shorthand `var(--rs-space-3) var(--rs-space-7) var(--rs-space-3) var(--rs-space-5)` so the right and left edges use the correct horizontal spacing tokens while top/bottom remain unchanged.

## Test plan
- [ ] Visual check: data-table toolbar in Storybook/playground matches Figma horizontal spacing (left `--rs-space-5`, right `--rs-space-7`).
- [ ] Verify no regressions in existing data-table stories (toolbar alignment, search input, action buttons).
- [ ] Confirm vertical padding (`--rs-space-3`) is unchanged.

### DIFF
--- a/packages/raystack/components/data-table/data-table.module.css
+++ b/packages/raystack/components/data-table/data-table.module.css
@@ -1,5 +1,5 @@
 .toolbar {
-  padding: var(--rs-space-3);
+  padding: var(--rs-space-3) var(--rs-space-7) var(--rs-space-3) var(--rs-space-5);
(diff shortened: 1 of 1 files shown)


rohanchkrabrty:

PR #778, opened 2026-05-07
fix(alert-dialog): remove footer border and default showCloseButton to false

rohanchkrabrty:
## Summary

- Removes the footer border on `AlertDialog.Body` via a new local `alert-dialog.module.css` (overrides the inherited `dialog` body `border-bottom`).
- **BREAKING:** flips the default of `AlertDialog.Content`'s `showCloseButton` prop from `true` to `false`. Existing consumers relying on the close button rendering by default must now opt in by passing `showCloseButton`. Please call this out in the changelog / release notes.
- Updates the docs prop table (`@default false`) and the existing test to explicitly opt into `showCloseButton` so test coverage of the close button is preserved.

## Test plan

- [ ] `pnpm test` (alert-dialog suite) passes
- [ ] Verify visually in the playground that `AlertDialog.Body` no longer has a bottom border
- [ ] Verify that `AlertDialog.Content` renders without a close button by default and that `<AlertDialog.Content showCloseButton>` still renders one
- [ ] Confirm the breaking change is documented in the changelog / release notes before publishing

### DIFF
--- a/apps/www/src/content/docs/components/alert-dialog/props.ts
+++ b/apps/www/src/content/docs/components/alert-dialog/props.ts
@@ -12,7 +12,7 @@ export interface AlertDialogContentProps {
 
   /**
    * Controls whether to show the close button
-   * @default true
+   * @default false
    */
   showCloseButton?: boolean;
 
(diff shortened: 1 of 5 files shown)


rohanchkrabrty:

PR #780, opened 2026-05-07
fix(combobox): hide placeholder when items selected in multi-select

rohanchkrabrty:


rohanchkrabrty:

PR #781, opened 2026-05-07
fix(sidebar): align spacing and tooltip offset with design

rohanchkrabrty:


rohanchkrabrty:

PR #785, opened 2026-05-07
fix(alert-dialog): remove footer border and default showCloseButton to false

rohanchkrabrty:
## Summary
- Dialog's body→footer separator moved from `.body { border-bottom }` to `:where(.body) + .footer { border-top }` (visually identical; the `:where()` keeps the rule's specificity at `.footer` so consumers can override cleanly).
- AlertDialog overrides this with a plain `.footer { border-top: 0 }` — no `:has()`, no class-on-body trick.
- **BREAKING:** `showCloseButton` default flipped from `true` to `false`. Existing consumers relying on the default close button will need to set `showCloseButton` explicitly — worth a changelog/release-notes callout.

### DIFF
--- a/packages/raystack/components/alert-dialog/alert-dialog-content.tsx
+++ b/packages/raystack/components/alert-dialog/alert-dialog-content.tsx
@@ -20,7 +20,7 @@ export interface AlertDialogContentProps
 export const AlertDialogContent = ({
   className,
   children,
-  showCloseButton = true,
+  showCloseButton = false,
   overlay,
   width,
   style,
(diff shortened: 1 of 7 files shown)


packages/raystack/components/alert-dialog/alert-dialog-content.tsx
```diff
@@ -20,7 +20,7 @@ export interface AlertDialogContentProps
 export const AlertDialogContent = ({
   className,
   children,
-  showCloseButton = true,
+  showCloseButton = false,
```
Shreyag02, line 23: Add the default value in docs and examples as well

Shreyag02 (changes requested):

Shreyag02 (approved):

PR #793, opened 2026-05-07
feat(amount): added bigInt and hideCurrency support in amount component

Shreyag02:
## Description

 Adds two capabilities to the Amount component:                                                    
                                                                                                                                                 
  1. BigInt support. value now accepts bigint in addition to `number | string`. `BigInt` is treated as integer-only and always in major units, so `valueInMinorUnits` is ignored when value is a `BigInt`. This restores precision for amounts beyond `Number.MAX_SAFE_INTEGER (≈ 9 × 10¹⁵) ` — the  previous workaround was string-only, which is awkward for backends that already produce `BigInt. Intl.NumberFormat.prototype.format()` accepts  `BigInt` natively, so the integration is direct.
  2. `hideCurrency` prop. A boolean that renders only the formatted number, with locale-driven separators, grouping, and currency-driven fraction digits intact. Useful for places where the currency is displayed elsewhere (column headers, line-item subtotals, token counters in the SDK tokens view). When hideCurrency is true, currencyDisplay is ignored.
  
Behavior contract documented in JSDoc on both source and docs props.ts. 
Backwards compatible — `hideCurrency` defaults to false, the `BigInt` branch is purely additive, and the negative-string path was previously buggy (no test ever passed against it).

## Type of Change

- [ ] Documentation update

## How Has This Been Tested?

  - Unit tests: added 11 new cases covering `BigInt (6)`, `hideCurrency (4)`, and the `negative-string regression (1)`. Full suite at 41/41 passing.   
    - BigInt: defaults, valueInMinorUnits=false parity, precision beyond `MAX_SAFE_INTEGER`, `negative BigInts`, `zero-decimal currency (JPY)`, no `MAX_SAFE_INTEGER` warning for `BigInt`.                                                                                                           
    - `hideCurrency`: basic case, BigInt input, override of currencyDisplay, currency-driven decimals still respected.
    - Negative string: value="-1299" → -$12.99.                                                                                                  
  - Type checks: zero TS diagnostics across all five files.                                                                                      
  - Manual trace: walked through edge-case combinations not directly tested — all produce correct output.
    ```<Amount value={1299n} hideCurrency hideDecimals />
    <Amount  value="-1" /> (padStart-padded to -$0.01)
    <Amount value={0n} />
    <Amount value={1299n} currency='BHD' /> 
  - Docs preview: added `hideCurrencyDemo` and a `BigInt` entry in `largeNumbersDemo`. The console-warning demo value was bumped from 15 digits (which never tripped Math.abs > MAX_SAFE_INTEGER) to 17 digits, so the warning now actually fires when the section is rendered.      

## Checklist:

N/A

## Related Issues

N/A


packages/raystack/components/amount/amount.tsx
```diff
@@ -171,37 +186,100 @@ export const Amount = ({
...
+      finalBaseValue.replace(/\D/g, '').length > 15
+    ) {
+      console.warn(
+        'Amount: a string longer than 15 digits is being formatted. ' +
+          'Older runtimes without Intl.NumberFormat V3 string-precision support ' +
+          '(Chrome <106, Firefox <116, Safari <15.4, Node <19) coerce strings via ' +
+          'Number() first, which loses precision beyond 2^53. Pass a bigint for ' +
+          'exact integer formatting if your targets include those runtimes.'
+      );
+    }
```
rohanchkrabrty: On more thought, I think this can be skipped. This issue might happens on older targets which exists around ~2022

We are anyways not supporting such older targets, but this warning will be present in all current usages making it a noise.

@paanSinghCoder what do you say?

packages/raystack/components/amount/amount.tsx
rohanchkrabrty, line 1:
```
 <Amount value='12.99' valueInMinorUnits /> // NaN
 <Amount value={12.99} valueInMinorUnits /> //Working
```
@Shreyag02 can you check this?

rohanchkrabrty (changes requested):

packages/raystack/components/amount/amount.tsx
paanSinghCoder: Yes, we can remove it. It creates noise.

paanSinghCoder:
@Shreyag02 Shouldn't this PR title be a `feat` instead of `fix`?

apps/www/src/content/docs/components/amount/demo.ts
```diff
@@ -149,13 +164,24 @@ export const withTextDemo = {
...
-    
-    {/* Numbers exceeding safe integer limit will show warning in console */}
-    <Amount value={999999999999999} /> {/* Will show warning */}
+    <Amount value="10000100091636935" 
+    valueInMinorUnits={false} hideDecimals />{/* $10,000,100,091,636,935 */}
+
+    {/* 
+    BigInt is always treated as major units — valueInMinorUnits is ignored 
+  */}
+    <Amount value={BigInt("9999999999999999999")} valueInMinorUnits={false} />{/* $9,999,999,999,999,999,999.00 */}
```
paanSinghCoder: `valueInMinorUnits={false}` can be removed from here as its contradicting. `valueInMinorUnits` is ignored if large value is passed to value prop.

apps/www/src/content/docs/components/amount/props.ts
```diff
@@ -1,14 +1,18 @@
 export interface AmountProps {
```
paanSinghCoder: `AmountProps` is duplicated in in `amount.tsx`. I think we can export it from there and use it here.

packages/raystack/components/amount/amount.tsx
paanSinghCoder: @Shreyag02 There is also a test case for this - "warns when a string longer than 15 digits is formatted"

Shreyag02: removed

apps/www/src/content/docs/components/amount/props.ts
Shreyag02: now imported from component

packages/raystack/components/amount/amount.tsx
Shreyag02: updated

apps/www/src/content/docs/components/amount/demo.ts
Shreyag02: removed

paanSinghCoder (approved):

rohanchkrabrty (approved):

PR #800, opened 2026-05-11
feat: indicator classNames and sidebar overflow

rohanchkrabrty:
## Summary

- Add `classNames` slot prop to `Indicator` (currently exposes the `indicator` slot) so consumers can target the inner badge without relying on the bare `className` pass-through.
- Update the Indicator docs (`props.ts`) to replace the removed `className` entry with the new `classNames` slot map.
- Remove a stray overflow rule from `sidebar.module.css` to fix layout bleed.

### DIFF
--- a/apps/www/src/content/docs/components/indicator/props.ts
+++ b/apps/www/src/content/docs/components/indicator/props.ts
@@ -14,6 +14,9 @@ export interface IndicatorProps {
   /** The content to show the indicator on */
   children: React.ReactNode;
 
+  /** Additional CSS class names */
+  className?: string;
+
   /** Map of classNames for internal components */
   classNames?: {
     /** Class for the indicator badge/dot element */
--- a/apps/www/src/content/docs/components/indicator/props.ts
+++ b/apps/www/src/content/docs/components/indicator/props.ts
@@ -19,7 +19,7 @@ export interface IndicatorProps {
 
   /** Map of classNames for internal components */
   classNames?: {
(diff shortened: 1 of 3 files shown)


apps/www/src/content/docs/components/indicator/props.ts
```diff
@@ -14,6 +14,9 @@ export interface IndicatorProps {
   /** The content to show the indicator on */
   children: React.ReactNode;
 
-  /** Additional CSS class names */
-  className?: string;
+  /** Map of classNames for internal components */
+  classNames?: {
```
paanSinghCoder, line 21: Breaking change

paanSinghCoder (approved):

apps/www/src/content/docs/components/indicator/props.ts
rohanchkrabrty: Updated the classNames to add support for Container. Not breaking anymore.

apps/www/src/content/docs/components/indicator/props.ts
```diff
@@ -16,4 +16,10 @@ export interface IndicatorProps {
 
   /** Additional CSS class names */
   className?: string;
+
+  /** Map of classNames for internal components */
+  classNames?: {
+    /** Class for the outer container element */
+    container?: string;
+  };
```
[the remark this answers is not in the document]
rohanchkrabrty: Both are supported

PR #810, opened 2026-05-15
fix: filter chip and DataTable.Filters layout

rohanchkrabrty:
## Summary

- Fix filter chip overflow: add `flex-shrink: 0` to label, operation, and remove button; allow the value field to shrink with `min-width: 0` + `overflow: hidden`, enabling ellipsis on long select values and inputs.
- Set `min-width: min-content` and `overflow: clip` on the chip itself so it never collapses below its un-shrinkable items.
- Simplify `DataTable.Filters` rendering: drop the inner `Flex` wrapper and the now-unused `classNames.container` key; chips render directly inside the outer `Flex` with `flex-wrap: wrap` for narrow toolbars.
- Toolbar now aligns to `start` with `gap={3}`, and the `AddFilter` icon button is sized `3` to match the chip height.
- Document the `classNames.container` removal in `docs/V1-migration.md`.

### DIFF
--- a/packages/raystack/components/filter-chip/filter-chip.module.css
+++ b/packages/raystack/components/filter-chip/filter-chip.module.css
@@ -8,15 +8,17 @@
   border-radius: var(--rs-radius-2);
   text-wrap: nowrap;
   width: fit-content;
+  min-width: min-content;
+  overflow: clip;
 }
 
 .chip-default {
   border: 1px solid var(--rs-color-border-base-tertiary);
-  box-shadow: var(--rs-shadow-feather);
 }
 
 .chip-label {
   padding: 0 var(--rs-space-3);
+  flex-shrink: 0;
 }
 .operationText {
   color: var(--rs-color-foreground-base-secondary);
@@ -26,6 +28,7 @@
   letter-spacing: var(--rs-letter-spacing-small);
   display: flex;
   align-items: center;
+  flex-shrink: 0;
 }
 
 .operationText:hover {
@@ -40,6 +43,7 @@
   background: transparent;
   min-height: unset;
   padding: 0;
+  text-overflow: ellipsis;
 }
 
 /* Control height of Select items */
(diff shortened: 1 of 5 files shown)


packages/raystack/components/filter-chip/filter-chip.module.css
```diff
@@ -102,19 +108,18 @@
...
 .selectValue * {
   font-weight: var(--rs-font-weight-regular);
+  min-width: 0;
 }
 
 .operationValue {
   color: var(--rs-color-foreground-base-secondary);
-}
-
-.selectColumn {
```
paanSinghCoder, line 115: Check if this className still has reference in the `filter-chip.tsx`. If yes, let's remove it.

packages/raystack/components/filter-chip/filter-chip.module.css
```diff
@@ -151,23 +157,24 @@ button.selectValue:hover {
...
-  border-radius: var(--rs-radius-1);
+  border-radius: 0;
   display: flex;
   justify-content: space-between;
   align-items: center;
-  width: 100%;
-  height: 18px;
-  min-height: 18px;
+  width: fit-content;
+  height: 22px;
```
rsbh, line 176: can we use apsara variable here

packages/raystack/components/filter-chip/filter-chip.module.css
```diff
@@ -132,14 +137,15 @@ button.selectValue:hover {
...
+  overflow: hidden;
 }
 
 .inputField {
-  border-radius: var(--rs-radius-1);
-  min-height: 18px;
-  height: 18px;
+  border-radius: 0;
+  min-height: 22px;
+  height: 22px;
```
rsbh, line 148: use apsara variable here

packages/raystack/components/filter-chip/filter-chip.module.css
rohanchkrabrty: We don't have variables defined for this

packages/raystack/components/filter-chip/filter-chip.module.css
rohanchkrabrty: We don't have variables defined for this

rsbh (approved):

paanSinghCoder (approved):

PR #812, opened 2026-05-18
fix: sidebar spacing and group margin

rohanchkrabrty:
## Summary
- Add `gap: var(--rs-space-6)` on sidebar root and remove the gap from `.main` so layout regions space themselves consistently.
- Switch `.header` padding to `0 var(--rs-space-4)` to align with the new root gap.
- Set `SidebarMain` to `gap={3}` so items inside main use 8px spacing.
- Add `margin-top: var(--rs-space-4)` (12px) to `.nav-group` inside `.main`, skipping the first child so a leading group sits flush at the top.

### DIFF
--- a/apps/www/src/content/docs/components/sidebar/demo.ts
+++ b/apps/www/src/content/docs/components/sidebar/demo.ts
@@ -23,7 +23,7 @@ export const preview = {
   code: sidebarLayout(`
   <Sidebar defaultOpen>
     <Sidebar.Header>
-      <Flex align="center" gap={3}>
+      <Flex align="center" gap={3} style={{padding:"4px"}}>
         <IconButton size={4} aria-label="Logo">
           <BellIcon width={24} height={24} />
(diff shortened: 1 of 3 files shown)


paanSinghCoder (approved):
Not related to this PR - Lets add an example for simple Sidebar without Group.

apps/www/src/content/docs/components/sidebar/demo.ts
```diff
@@ -23,7 +23,7 @@ export const preview = {
   code: sidebarLayout(`
   <Sidebar defaultOpen>
     <Sidebar.Header>
-      <Flex align="center" gap={3}>
+      <Flex align="center" gap={3} style={{padding:"4px"}}>
```
paanSinghCoder: Replace with var

rohanchkrabrty: Done

PR #816, opened 2026-05-20
feat(color-picker): migrate to culori and add add OKLCH support

paanSinghCoder:
## Summary

- Swaps `color` for `culori` in the `ColorPicker` component so it can accept and emit `oklch()` alongside hex/rgb/hsl
- Adds `oklch` to `SUPPORTED_MODES` and a `formatOklch()` helper (4-decimal L/C, 2-decimal H, hue pinned to 0 for grays)
- Updates docs (props, demo, mdx) and adds an example page at `/examples/color-picker`
- Add copy-to-clipboard button

## Why culori (not color-convert)

The picker's `value` prop accepts any CSS color string from consumers — hex, rgb/rgba, hsl/hsla, oklch, named colors. `color-convert` is route-based; it has no CSS parser and no stringifiers (except a hex helper), so we'd hand-roll ~170 lines of parsers and formatters. `culori` ships exactly the surface this component needs.

| culori util used here | What it does | color-convert equivalent? |
| --- | --- | --- |
| `parse` | Any CSS string → typed color object | No — color-convert takes numeric tuples only |
| `converter('hsl')` | Any color object → HSL | No — explicit routes only (`c.rgb.hsl`, `c.hex.rgb`, …) |
| `formatHex` | Color → `#rrggbb` | Partial — `c.rgb.hex` returns `'FF0000'` (no `#`, no alpha) |
| `formatHex8` | Color → `#rrggbbaa` | No — manual alpha-to-hex concat |
| `formatHsl` | Color → `hsl(...)` / `hsla(...)` | No — manual template |
| `formatRgb` | Color → `rgb(...)` / `rgba(...)` | No — manual template |

## Caveat

Sliders still operate in HSL. `oklch` is an I/O format, not a perceptual editing mode — an `oklch()` value passed in will not be bit-identical when read back, because the round-trip goes through HSL → sRGB → OKLCH.

## Follow-up

The VS Code plugin (#759) currently uses `color-convert` because its input is a closed set of `oklch()` tokens. We could move it to `culori` later to consolidate on one color library across the monorepo — the trade-off is ~9 KB gzipped in the plugin bundle vs. removing the parser/scaling/type-shim glue we hand-rolled there.

## Test plan

- [x] `pnpm test components/color-picker` — 24/24 passing (3 new tests for oklch input, oklch mode output, and oklch alpha tail)
- [x] `pnpm tsc --noEmit` clean on touched files
- [x] `pnpm biome check` clean on touched files
- [ ] Visit `/examples/color-picker` in the docs site and verify each card

### DIFF
--- a/apps/www/src/content/docs/components/color-picker/demo.ts
+++ b/apps/www/src/content/docs/components/color-picker/demo.ts
@@ -16,7 +16,7 @@ style={{
 <ColorPicker.Alpha />
 <Flex direction="row" gap={2}>
   <ColorPicker.Mode />
-  <ColorPicker.Input />
+  <ColorPicker.Input copyable />
 </Flex>
 </ColorPicker>`
 };
@@ -36,6 +36,23 @@ export const basicDemo = {
 `
 };
 
+export const copyableDemo = {
+  type: 'code',
+  code: `<ColorPicker defaultValue='#5B8DEF' style={{
+  width: '240px',
+  height: '300px',
+  padding: 12,
+  background: 'white'
+}}>
+  <ColorPicker.Area />
+  <ColorPicker.Hue />
+  <Flex direction='row' gap={2}>
+    <ColorPicker.Mode />
+    <ColorPicker.Input copyable />
+  </Flex>
+</ColorPicker>`
+};
+
 export const oklchDemo = {
   type: 'code',
   code: `<ColorPicker
@@ -53,7 +70,7 @@ style={{
 <ColorPicker.Alpha />
 <Flex direction="row" gap={2}>
   <ColorPicker.Mode />
-  <ColorPicker.Input />
+  <ColorPicker.Input copyable />
 </Flex>
 </ColorPicker>`
 };
@@ -90,7 +107,7 @@ export const popoverDemo = {
           <ColorPicker.Alpha />
           <Flex direction='row' gap={2}>
             <ColorPicker.Mode />
-            <ColorPicker.Input />
+            <ColorPicker.Input copyable />
           </Flex>
         </ColorPicker>
       </Popover.Content>
--- a/apps/www/src/content/docs/components/color-picker/demo.ts
+++ b/apps/www/src/content/docs/components/color-picker/demo.ts
@@ -36,6 +36,28 @@ export const basicDemo = {
 `
 };
 
+export const oklchDemo = {
+  type: 'code',
+  code: `<ColorPicker
+defaultValue='oklch(0.5438 0.191 267.01)'
+defaultMode='oklch'
+style={{
+  width: '240px',
+  height: '320px',
+  padding: 12,
+  background: 'white'
+}}
+>
+<ColorPicker.Area />
+<ColorPicker.Hue />
+<ColorPicker.Alpha />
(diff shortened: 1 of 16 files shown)


packages/raystack/components/color-picker/utils.ts
```diff
@@ -1,17 +1,13 @@
...
-  else if (mode === 'hsl') string = color.hsl().toString();
-  else string = color.rgb().toString();
-
-  return string;
-};
-
-export const SUPPORTED_MODES = ['hex', 'hsl', 'rgb'];
+export const SUPPORTED_MODES = ['hex', 'hsl', 'rgb', 'oklch'];
 
 export type ModeType = (typeof SUPPORTED_MODES)[number];
```
Shreyag02, line 14: We should use Union of Literals for the mode type

Shreyag02 (approved):

packages/raystack/components/color-picker/utils.ts
paanSinghCoder: Done

Shreyag02 (approved):

apps/www/src/content/docs/components/color-picker/demo.ts
```diff
@@ -36,6 +36,45 @@ export const basicDemo = {
...
+</ColorPicker>`
+};
+
+export const oklchDemo = {
+  type: 'code',
+  code: `<ColorPicker
+defaultValue='oklch(0.5438 0.191 267.01)'
+defaultMode='oklch'
+style={{
+  width: '240px',
```
Shreyag02, line 62: Moving only the slider values leaves the first component of the OKLCH value unchanged, making it look like the value isn't updating. Since the cursor also resets upon updating the slider, consider increasing the card's width to resolve this.

Shreyag02 (approved):

apps/www/src/content/docs/components/color-picker/demo.ts
paanSinghCoder: Noted. I think this is the tradeoff we can live with for now. Will point this with design and lets raise a different PR to increase the width.

packages/raystack/components/color-picker/color-picker-alpha.tsx
```diff
@@ -11,7 +12,17 @@ export const ColorPickerAlpha = ({
...
+  const { alpha = 1, lightness, chroma, hue, setColor } = useColorPicker();
+  // Drive the track gradient from the current color so the user can preview
+  // what the picked tone looks like at any alpha. The CSS reads
+  // --rs-color-picker-alpha-end from this style.
+  const trackStyle = useMemo(
+    () =>
+      ({
+        ['--rs-color-picker-alpha-end' as string]: `oklch(${lightness} ${chroma} ${hue})`
+      }) as CSSProperties,
+    [lightness, chroma, hue]
```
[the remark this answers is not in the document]
paanSinghCoder: Intentional. This is a render-only CSS value for the gradient endpoint. It must stay alpha-free, so `formatOklchString` would be wrong here since it appends / alpha. Hue is already finite from `parseColor`, and full precision is fine for a CSS paint value.

PR #817, opened 2026-05-21
chore: forward props in drawer header

rohanchkrabrty:
## Summary

- Spread native `div` props onto `Drawer.Header` so callers can pass `id`, `role`, `aria-*`, event handlers, etc.
- Replace the bespoke `{ children, className }` prop type with `ComponentProps<'div'>` for full HTML compatibility.
- Drop the now-unused `ReactNode` import.

### DIFF
--- a/packages/raystack/components/spinner/spinner.module.css
+++ b/packages/raystack/components/spinner/spinner.module.css
@@ -1,5 +1,4 @@
 .spinner {
-  display: inline-block;
   position: relative;
   display: flex;
   justify-content: center;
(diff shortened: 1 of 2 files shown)


packages/raystack/components/spinner/spinner.module.css
```diff
@@ -1,5 +1,4 @@
 .spinner {
-  display: inline-block;
   position: relative;
   display: flex;
```
paanSinghCoder, line 3: `display: flex;` and `justify-content: center;` are also not required as the .spinner only renders 8 poles that are absolutely positioned. Only css required here is `position: relative;`. Please confirm once.

rohanchkrabrty: It's needed because the the sub items are aligned at center and then transformed against.

paanSinghCoder (approved):

PR #842, opened 2026-07-02
feat: guided tour component

rohanchkrabrty:
## Summary

- Add the guided **Tour** component (RFC 003) for building step-by-step onboarding flows, built on Base UI Popover with a spotlight overlay
- Split the implementation into composable parts (`tour-root`, `tour-content`, `tour-overlay`, `tour-parts`, `tour-context`) with supporting hooks (`use-tour-target`, `use-prefers-reduced-motion`) and shared types
- Export the component from the package root and add full test coverage for stepping, targeting, and overlay edge cases
- Wire up docs and playground: MDX guide, props reference, live demo, and interactive tour examples on the www site

### DIFF
--- a/packages/raystack/components/tour/tour-overlay.tsx
+++ b/packages/raystack/components/tour/tour-overlay.tsx
@@ -1,0 +1,196 @@
+'use client';
+
+import { cx } from 'class-variance-authority';
+import { type ComponentPropsWithoutRef, useEffect, useState } from 'react';
+import { createPortal } from 'react-dom';
+import styles from './tour.module.css';
+import { useTourContext } from './tour-context';
+
+export interface TourOverlayProps extends ComponentPropsWithoutRef<'div'> {
+  /**
+   * Space between the target and the spotlight edge in pixels; steps can
+   * override. @default 4
+   */
+  spotlightPadding?: number;
+  /** Spotlight corner radius in pixels; steps can override. @default 6 */
+  spotlightRadius?: number;
+  /**
+   * Allow pointer interaction with the spotlighted element; steps can
+   * override. @default false
+   */
+  spotlightClicks?: boolean;
+}
+
+interface SpotlightRect {
+  x: number;
+  y: number;
+  width: number;
+  height: number;
+}
+
+function rectsEqual(a: SpotlightRect, b: SpotlightRect) {
+  return (
+    Math.abs(a.x - b.x) < 0.5 &&
+    Math.abs(a.y - b.y) < 0.5 &&
+    Math.abs(a.width - b.width) < 0.5 &&
+    Math.abs(a.height - b.height) < 0.5
+  );
+}
+
+/**
+ * Tracks an element's viewport rect with a frame loop, so the spotlight
+ * follows through scrolling (any container), resizes and CSS animations — e.g.
+ * a dialog's entry transform, which fires no scroll/resize events. Re-renders
+ * only when the rect actually changes.
+ *
+ * When the element changes the previous rect is intentionally kept until the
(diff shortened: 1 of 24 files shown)


packages/raystack/components/tour/tour-overlay.tsx
```diff
@@ -0,0 +1,280 @@
...
+  y: number;
+  width: number;
+  height: number;
+}
+
+/**
+ * Whether two rects match within half a pixel on every side. Shared with the
+ * root's reveal loop, which feeds it `DOMRect`s (`x`/`y` mirror `left`/`top`).
+ */
+export function rectsEqual(a: SpotlightRect, b: SpotlightRect) {
```
Shreyag02: Move rectsEqual (and SpotlightRect) out of the overlay into a shared utils.ts.
Right now rectsEqual lives in the overlay component but tour-root.tsx imports it — so the root is reaching into a leaf component to grab geometry helpers. 

That's backwards: the parent shouldn't depend on internals of a child.

packages/raystack/components/tour/tour-root.tsx
```diff
@@ -0,0 +1,476 @@
...
+    if (!open) {
+      lastActiveIndexRef.current = -1;
+      return;
+    }
+    if (!popoverOpen || !step || lastActiveIndexRef.current === index) return;
+    lastActiveIndexRef.current = index;
+    emit({ type: 'step:active', index, step });
+  }, [open, popoverOpen, index, step, emit]);
+
+  // Scroll the target into view once, when the step becomes active.
```
Shreyag02: Since step is an object rebuilt on every render (all our examples pass steps inline), this effect fires on every host re-render — and if the target is out of view, it calls scrollIntoView again, moving the page back while the user is scrolling mid-step.

should swap step for index in the deps (keep anchor so late-resolving targets still scroll in).

packages/raystack/components/tour/use-tour-target.ts
```diff
@@ -0,0 +1,134 @@
...
+      return;
+    }
+    if (target == null) {
+      // Detached step: nothing to resolve, the popover renders centered.
+      setElement(null);
+      setState('found');
+      return;
+    }
+
+    let timer: ReturnType<typeof setTimeout> | undefined;
```
Shreyag02, line 57: The timer slot is cleared before the callback runs, so if unrelated DOM churn triggers another reconcile() while the target is still missing, a fresh timeout gets scheduled and onNotFound fires again — meaning a possible double skip/stop.

packages/raystack/components/tour/use-tour-target.ts
```diff
@@ -0,0 +1,134 @@
...
+    } catch {
+      // A throwing resolver (e.g. `() => editor.getNode('x')` before the editor
+      // is ready) should degrade to "not found", not crash the tour — matching
+      // how an invalid selector is handled above.
+      return null;
+    }
+  }
+  if (target instanceof Element) return target;
+  return target.current ?? null;
+}
```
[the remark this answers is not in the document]
Shreyag02: should add the suggested fix

Shreyag02 (changes requested):

packages/raystack/components/tour/use-tour-target.ts
rohanchkrabrty: This shouldn't fire twice in practice. When the timeout triggers, `onNotFound` advances the tour or closes it, which changes the effect's dependencies and immediately cleans up the observer and timer.

Shreyag02: got it

Shreyag02 (approved):

PR #852, opened 2026-07-11
fix: bring component motion in line with animation standards

ravisuhag:
## Summary

An animation audit across the component library found a set of motion issues, all now fixed:

- Consolidated hand-typed easing curves/durations into shared `--rs-ease-*` / `--rs-duration-*` tokens (`styles/effects.css`)
- Removed the command palette's open/close animation — keyboard-driven, used 100+ times/day, should never animate
- Rebuilt popover/menu/select/combobox entrances as interruptible, origin-anchored transitions (were keyframes with no exit, or no entrance at all)
- Fixed a toast transition that fought the user's active swipe-to-dismiss gesture
- Replaced `transition: all` with the interactive token in button/select/field/text-area (was animating unintended layout properties)
- Synced switch/checkbox/radio color transitions with their movement (color used to snap while the thumb/mark animated)
- Fixed an inverted `prefers-reduced-motion` override in the sidebar (reduced-motion users were getting *more* animation than everyone else)
- Moved the tabs indicator off `top`/`left` onto `translate` (was triggering layout on every tab switch)
- Added missing entrance motion to checkbox, radio, and empty-state
- Tokenized dialog/tour/skeleton/progress/meter durations and easings; removed a dead unused keyframe

Checkbox/radio also needed a `keepMounted` addition on their Base UI indicators so the exit fade can actually play (the indicator previously unmounted synchronously on uncheck, before any transition could run); the check icon's render logic was adjusted so it doesn't leave stray hidden markup in the DOM at rest.

## Test plan

- [x] `pnpm build:apsara` passes
- [x] `pnpm test:apsara` — 1794 passed, 1 skipped; 7 test files fail on a pre-existing `@base-ui` module-resolution issue unrelated to this change (confirmed by reproducing the same failure on unmodified `main`)

### DIFF
--- a/packages/raystack/components/button/button.module.css
+++ b/packages/raystack/components/button/button.module.css
@@ -210,7 +210,7 @@
 
 @media (prefers-reduced-motion: no-preference) {
   .button {
-    transition: all 0.2s ease;
+    transition: var(--rs-transition-interactive);
   }
 }
 
--- a/packages/raystack/components/checkbox/checkbox.module.css
+++ b/packages/raystack/components/checkbox/checkbox.module.css
@@ -20,6 +20,7 @@
   border: 1px solid var(--rs-color-border-base-secondary);
   cursor: pointer;
   flex-shrink: 0;
+  transition: background-color 150ms ease, border-color 150ms ease;
 }
 
 /* Size variants */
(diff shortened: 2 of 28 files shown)


ravisuhag:
Superseded by #856, which consolidates the animation work and the design/interaction pass (with review changes folded in) into one branch.

Issue #642, opened 2026-02-18
[Table] Add displayName, scope attribute support, and caption element


ravisuhag:
## Summary

Table is a custom semantic HTML table wrapper using CVA + CSS Modules. Missing displayName on all sub-components, accessibility attributes, and API completeness.

### 1. Add `displayName` to all sub-components
None of TableRoot, TableHeader, TableBody, TableRow, TableHead, TableCell, or SectionHeader have displayName set. All other compound components in the library have this.

### 2. Add `scope` attribute support to TableHead (WCAG 2.1)
`<th>` elements should support `scope="col"` or `scope="row"` to programmatically associate headers with data cells.

### 3. Add `<caption>` element support
No `Caption` sub-component exposed for screen reader context, despite CSS having `caption-side: bottom`.

### 4. Add `aria-sort` support for sortable columns
Headers can't indicate sort direction, needed for sortable table patterns.

### 5. Replace hardcoded CSS values
- `border-bottom: 0.5px` — sub-pixel rendering issue
- Sticky header `top: 0px` — should be `top: 0`

### 6. Normalize type signatures across sub-components
TableHead and TableCell use CVA variants but TableBody and TableRow don't. Inconsistent API surface.

### 7. Handle text truncation accessibility
Cells use `text-overflow: ellipsis` which hides content with no tooltip or accessible overflow mechanism.

## Files
- `packages/raystack/raystack/components/table/`

ravisuhag:
Partially addressed by PR #856 (commit `96e9ee9d`).

- **Item 6 (normalize type signatures)** — `TableRow` now uses a CVA `row` variant, and the Table primitive gained row-state hooks (selected / pressed). This closes part of the API-consistency gap between sub-components.

Still open:
- **Item 7 (truncation accessibility)** — cells still use `text-overflow: ellipsis` with no tooltip or accessible overflow mechanism.
- The sub-pixel border item was moved to the design-tokens cross-cutting issue (#672).

Keeping this open for the remaining items.

Issue #596, opened 2026-02-18
[AnnouncementBar] Improve accessibility, variants, and component API


ravisuhag:
## Summary

Improvements to the AnnouncementBar component to fix accessibility issues, add missing features, and improve consistency.

### 1. Add `forwardRef` and forward HTML attributes
The component doesn't use `forwardRef` and `AnnouncementBarProps` doesn't extend HTML attributes. For consistency with the rest of the design system, add `forwardRef` and extend the type to accept standard HTML/Flex props (`style`, `role`, `aria-label`, `id`, etc.).

### 2. Make action button a `<button>` (a11y)
The action area uses `<Text onClick={...}>` which is not keyboard accessible — it won't be focusable or activatable via Enter/Space. Should be a `<button>` or use the design system's `Button` component.

### 3. Add `dismissible` / `onDismiss` support
No way to dismiss/close the bar. A common pattern is an `onDismiss` callback with a close button.

### 4. Accept `ReactNode` for `text` prop
The `text` prop only accepts `string`. It could accept `ReactNode` for richer content (e.g., links, bold text).

### 5. Add `warning`, `success`, `info` variants
Only 3 variants exist (`normal`, `gradient`, `error`). Common additions: `warning`, `success`, `info`.

### 6. Replace hardcoded gradient with design tokens
The gradient variant uses hardcoded hex values (`#eeaeca`, `#df94e9`) instead of design tokens.

### 7. Fix `announementBar` typo
The CVA variable name `announementBar` is missing a `c` — should be `announcementBar`.

ravisuhag:
Partially addressed by PR #856 (commit `f5aa1348`).

- The action button now has interactive states — hover, active, and a `:focus-visible` ring — using the shared `--rs-transition-interactive` token.

Still open:
- **Item 2** — dismissible / `onDismiss` support.
- **Item 3** — accept `ReactNode` for the `text` prop.
- **Item 4** — `warning` / `success` / `info` variants.
- **Item 6** — the `announementBar` typo.

Keeping this open for the remaining items.

PR #856, opened 2026-07-12
fix: design-system audit — motion, interaction states, focus & accessibility

### LINKED ISSUE #621: [Image] Add forwardRef and remove redundant accessibility attributes

### LINKED ISSUE #625: [Link] Component improvements
## Summary

Link wraps the Text component as a native `<a>` element. Has type safety issues with aria-label generation and missing CSS states.

## Moved to Cross-Cutting Issues

The following items have been moved to cross-cutting issues that address them across all components:

- Replace hardcoded CSS values with design tokens (#672)
- Accessibility baseline (#673)
- Animation: prefers-reduced-motion (#677)
- Consistent focus-visible ring styles (#681)
- Standardize disabled state styling (#680)

## Files
- `packages/raystack/raystack/components/link/link.tsx`
- `packages/raystack/raystack/components/link/link.module.css`

ravisuhag:
## Summary

A design-system-wide audit of the component library, consolidating the animation work and the broader interaction/craft pass into a single branch.

### What's covered

- **Motion** — shared easing/duration/scale tokens; interruptible, origin-anchored entrances for overlays (popover, menu, tooltip, select, combobox, preview-card); indeterminate Progress; drawer 1:1 swipe tracking; reduced-motion fallbacks throughout.
- **Interaction states** — hover / press / focus feedback for buttons, toggles, chips, and the text-field family (input, field, number-field, text-area, otp-field), with a unified invalid state.
- **Focus** — a consistent, tokenized `:focus-visible` ring across the library.
- **Accessibility** — reduced-motion and reduced-transparency fallbacks; a calendar day can no longer be selected by keyboard while the grid is loading; announcement-bar action-button states.
- **Data views** — selected / pressed row states, sort-icon direction, sticky-header hairlines that appear only once content scrolls under them, tabular figures.
- **Media** — image and avatar fade-in with a cached-image skip and a one-shot fallback guard.
- **Tokens & typography** — motion/interaction tokens consolidated into `styles/effects.css`; Inter body tracking and the h1–h4 type ramp retuned.

## Related issues

Closes #681 — consistent, tokenized `:focus-visible` ring applied across the library.
Closes #621 — Image error/loading handling: one-shot fallback guard plus load-state fade-in.
Closes #625 — Link's remaining items (focus-visible, reduced-motion) are now done.

Partially addresses:
- #642 — `TableRow` now uses CVA and gains row-state hooks. Cell truncation accessibility is still open.
- #596 — AnnouncementBar action-button interactive states added. Variants, dismissible support, and the `announementBar` typo are still open.

## Test plan

- [x] `pnpm build` passes.
- Component test suites pass. A pre-existing `@base-ui` module-resolution failure affects a handful of unrelated test files on `main` as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -222,10 +222,10 @@ Apsara follows an automated release process using GitHub Actions and semantic ve
    # Ensure you're on the correct branch
    git checkout main  # for production release
    # OR
-   git checkout develop  # for release candidate
+   git checkout release/x.y  # for release candidate
    
    # Pull latest changes
-   git pull origin main  # or develop
+   git pull origin main  # or the release branch
    ```
 
 2. **Create and push a tag**:
@@ -255,7 +255,7 @@ The release process includes these automated steps:
 3. **Install** dependencies
 4. **Build** the library using `pnpm ci:build`
 5. **Bump version** in package.json based on the git tag
-6. **Publish** to NPM using `release-it`
+6. **Publish** to NPM using `pnpm publish` (both `@raystack/apsara` and `@raystack/tools-config`)
 7. **Generate** GitHub release notes
 
 ### NPM Publishing
--- a/packages/raystack/components/accordion/accordion.module.css
+++ b/packages/raystack/components/accordion/accordion.module.css
@@ -31,9 +31,18 @@
   border-top: none;
 }
 
-.accordion-trigger:hover,
-.accordion-trigger:focus-visible {
+.accordion-trigger:hover {
   background-color: var(--rs-color-background-base-primary-hover);
+}
+
+/* Trigger sits flush in a clipped container: keep the ring inside. */
+.accordion-trigger:focus-visible {
+  outline: var(--rs-focus-ring);
+  outline-offset: var(--rs-focus-ring-offset-inset);
+}
+
+.accordion-trigger:active:not(:disabled) {
+  background-color: var(--rs-color-background-neutral-secondary);
 }
 
 .accordion-trigger:disabled {
@@ -80,10 +89,10 @@
 
 @media (prefers-reduced-motion: no-preference) {
   .accordion-icon {
-    transition: transform 150ms ease-in-out;
+    transition: transform var(--rs-duration-fast) var(--rs-ease-in-out);
   }
 
   .accordion-content {
-    transition: height 150ms ease-out;
+    transition: height var(--rs-duration-fast) var(--rs-ease-out);
   }
 }
--- a/packages/raystack/components/accordion/accordion.module.css
+++ b/packages/raystack/components/accordion/accordion.module.css
@@ -35,10 +35,11 @@
   background-color: var(--rs-color-background-base-primary-hover);
 }
 
-/* Trigger sits flush in a clipped container: keep the ring inside. */
+/* Trigger has its own border: hug it with the ring so they don't read as
+   two separate lines. */
 .accordion-trigger:focus-visible {
   outline: var(--rs-focus-ring);
-  outline-offset: var(--rs-focus-ring-offset-inset);
+  outline-offset: var(--rs-focus-ring-offset-inset-border);
 }
 
 .accordion-trigger:active:not(:disabled) {
--- a/packages/raystack/components/accordion/accordion.module.css
+++ b/packages/raystack/components/accordion/accordion.module.css
@@ -95,4 +95,8 @@
   .accordion-content {
     transition: height var(--rs-duration-fast) var(--rs-ease-out);
   }
+
+  .accordion-trigger {
+    transition: var(--rs-transition-interactive);
+  }
 }
--- a/packages/raystack/components/announcement-bar/announcement-bar.module.css
+++ b/packages/raystack/components/announcement-bar/announcement-bar.module.css
@@ -51,18 +51,29 @@
   font: inherit;
   color: inherit;
   border-radius: var(--rs-radius-1);
-  transition: var(--rs-transition-interactive);
 }
 
 .action-btn:hover {
   opacity: 0.8;
 }
 
+/* Press reads as scale, not a second dim — 0.64 text mid-press is barely
+   legible on the gradient variant, and hover's 0.8 still applies. */
 .action-btn:active {
-  opacity: 0.64;
+  transform: scale(var(--rs-scale-pressed));
 }
 
 .action-btn:focus-visible {
   outline: 1px solid currentColor;
   outline-offset: 2px;
+}
+
+@media (prefers-reduced-motion: no-preference) {
+  .action-btn {
+    transition: var(--rs-transition-interactive);
+  }
+
+  .action-btn:active {
+    transition: var(--rs-transition-pressed);
+  }
 }
--- a/packages/raystack/components/announcement-bar/announcement-bar.module.css
+++ b/packages/raystack/components/announcement-bar/announcement-bar.module.css
@@ -50,4 +50,19 @@
   padding: 0;
   font: inherit;
   color: inherit;
+  border-radius: var(--rs-radius-1);
+  transition: var(--rs-transition-interactive);
+}
+
+.action-btn:hover {
+  opacity: 0.8;
+}
+
+.action-btn:active {
+  opacity: 0.64;
+}
+
+.action-btn:focus-visible {
+  outline: 1px solid currentColor;
+  outline-offset: 2px;
 }
--- a/packages/raystack/components/avatar/avatar.tsx
+++ b/packages/raystack/components/avatar/avatar.tsx
@@ -1,8 +1,20 @@
+'use client';
+
 import { Avatar as AvatarPrimitive } from '@base-ui/react/avatar';
 import { cva, cx, VariantProps } from 'class-variance-authority';
-import { ComponentProps, isValidElement, ReactElement, ReactNode } from 'react';
+import {
+  ComponentProps,
+  isValidElement,
+  ReactElement,
+  ReactNode,
+  useRef,
+  useState
+} from 'react';
 import styles from './avatar.module.css';
 import { AVATAR_COLORS } from './utils';
+
+// Matches Base UI's AvatarRoot ImageLoadingStatus union.
+type ImageLoadingStatus = 'idle' | 'loading' | 'loaded' | 'error';
 
 const avatar = cva(styles.avatar, {
   variants: {
@@ -164,21 +176,42 @@ const AvatarRoot = ({
   variant,
   color,
   ...props
-}: AvatarProps) => (
-  <AvatarPrimitive.Root
-    className={cx(
-      styles.imageWrapper,
-      avatar({ size, radius, variant, color }),
-      className
-    )}
-    {...props}
-  >
-    <AvatarPrimitive.Image className={image()} src={src} alt={alt} />
-    <AvatarPrimitive.Fallback className={styles.fallback}>
-      {fallback}
-    </AvatarPrimitive.Fallback>
-  </AvatarPrimitive.Root>
-);
+}: AvatarProps) => {
+  const sawLoadingRef = useRef(false);
+  const [fadeIn, setFadeIn] = useState(false);
+  const handleLoadingStatusChange = (status: ImageLoadingStatus) => {
+    if (status === 'loading') {
+      sawLoadingRef.current = true;
+      setFadeIn(false);
+    } else if (status === 'loaded') {
+      // A cached image reports 'loaded' with no 'loading' phase (Base UI never
+      // emits it), so it stays instant; only network loads fade in.
+      setFadeIn(sawLoadingRef.current);
+      sawLoadingRef.current = false;
+    }
+  };
+
+  return (
+    <AvatarPrimitive.Root
+      className={cx(
+        styles.imageWrapper,
+        avatar({ size, radius, variant, color }),
+        className
+      )}
+      {...props}
+    >
+      <AvatarPrimitive.Image
+        className={cx(image(), fadeIn && styles['image-fade-in'])}
+        src={src}
+        alt={alt}
+        onLoadingStatusChange={handleLoadingStatusChange}
+      />
+      <AvatarPrimitive.Fallback className={styles.fallback}>
+        {fallback}
+      </AvatarPrimitive.Fallback>
+    </AvatarPrimitive.Root>
+  );
+};
 
 AvatarRoot.displayName = 'Avatar';
 
--- a/packages/raystack/components/breadcrumb/breadcrumb.module.css
+++ b/packages/raystack/components/breadcrumb/breadcrumb.module.css
@@ -50,9 +50,7 @@
 }
 
 .breadcrumb-link:focus-visible {
-  outline: none;
-  background-color: var(--rs-color-background-base-primary-hover);
-  color: var(--rs-color-foreground-base-secondary);
+  outline: var(--rs-focus-ring);
   border-radius: var(--rs-radius-1);
 }
 
@@ -95,8 +93,7 @@
 }
 
 .breadcrumb-dropdown-trigger:focus-visible {
-  background-color: var(--rs-color-background-base-primary-hover);
-  color: var(--rs-color-foreground-base-primary);
+  outline: var(--rs-focus-ring);
   border-radius: var(--rs-radius-1);
 }
 
--- a/packages/raystack/components/breadcrumb/breadcrumb.module.css
+++ b/packages/raystack/components/breadcrumb/breadcrumb.module.css
@@ -39,8 +39,21 @@
   cursor: pointer;
 }
 
+@media (prefers-reduced-motion: no-preference) {
+  .breadcrumb-link {
+    transition: color var(--rs-duration-fast) var(--rs-ease-out);
+  }
+}
+
 .breadcrumb-link:hover {
   color: var(--rs-color-foreground-base-secondary);
+}
+
+.breadcrumb-link:focus-visible {
+  outline: none;
+  background-color: var(--rs-color-background-base-primary-hover);
+  color: var(--rs-color-foreground-base-secondary);
+  border-radius: var(--rs-radius-1);
 }
 
 .breadcrumb-link-active {
@@ -79,6 +92,12 @@
   font-weight: inherit;
   outline: none;
   padding: 0;
+}
+
+.breadcrumb-dropdown-trigger:focus-visible {
+  background-color: var(--rs-color-background-base-primary-hover);
+  color: var(--rs-color-foreground-base-primary);
+  border-radius: var(--rs-radius-1);
 }
 
 .breadcrumb-dropdown-icon {
--- a/packages/raystack/components/calendar/date-picker.tsx
+++ b/packages/raystack/components/calendar/date-picker.tsx
@@ -208,20 +208,29 @@ export function DatePicker({
   }
 
   const defaultTrigger = (
-    <Input
-      size='small'
-      placeholder='Select date'
-      aria-invalid={!!error}
-      className={styles.datePickerInput}
-      trailingIcon={showCalendarIcon ? <CalendarIcon /> : undefined}
-      {...inputProps}
-      ref={popover.inputRef}
-      value={inputValue}
-      onChange={handleInputChange}
-      onFocus={popover.handleInputFocus}
-      onBlur={popover.handleInputBlur}
-      onKeyUp={handleKeyUp}
-    />
+    <>
+      <Input
+        size='small'
+        placeholder='Select date'
+        aria-invalid={!!error}
+        className={styles.datePickerInput}
+        trailingIcon={showCalendarIcon ? <CalendarIcon /> : undefined}
+        {...inputProps}
+        ref={popover.inputRef}
+        value={inputValue}
+        onChange={handleInputChange}
+        onFocus={popover.handleInputFocus}
+        onBlur={popover.handleInputBlur}
+        onKeyUp={handleKeyUp}
+      />
+      <span
+        className={styles.datePickerError}
+        data-visible={error ? '' : undefined}
+        role='alert'
+      >
+        {error}
+      </span>
+    </>
   );
 
   /*
--- a/packages/raystack/components/calendar/date-picker.tsx
+++ b/packages/raystack/components/calendar/date-picker.tsx
@@ -6,7 +6,7 @@ import dayjs from 'dayjs';
 import customParseFormat from 'dayjs/plugin/customParseFormat';
 import isSameOrAfter from 'dayjs/plugin/isSameOrAfter';
 import isSameOrBefore from 'dayjs/plugin/isSameOrBefore';
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useId, useRef, useState } from 'react';
 import { PropsBase } from 'react-day-picker';
 import { Input } from '../input';
 import { InputProps } from '../input/input';
@@ -207,6 +207,15 @@ export function DatePicker({
     }
   }
 
+  const errorId = useId();
+  // Link the error text to the input so a user tabbing into an already-erroring
+  // field hears the reason, not just "invalid" (role="alert" alone only fires
+  // when the message first appears). Merge with any consumer-supplied value.
+  const describedBy =
+    [inputProps['aria-describedby'], error ? errorId : undefined]
+      .filter(Boolean)
+      .join(' ') || undefined;
+
   const defaultTrigger = (
     <>
       <Input
@@ -216,6 +225,7 @@ export function DatePicker({
         className={styles.datePickerInput}
         trailingIcon={showCalendarIcon ? <CalendarIcon /> : undefined}
         {...inputProps}
+        aria-describedby={describedBy}
         ref={popover.inputRef}
         value={inputValue}
         onChange={handleInputChange}
@@ -224,6 +234,7 @@ export function DatePicker({
         onKeyUp={handleKeyUp}
       />
       <span
+        id={errorId}
         className={styles.datePickerError}
         data-visible={error ? '' : undefined}
         role='alert'
--- a/packages/raystack/components/callout/callout.tsx
+++ b/packages/raystack/components/callout/callout.tsx
@@ -6,11 +6,15 @@ import {
   type ComponentProps,
   type CSSProperties,
   type ReactNode,
+  useEffect,
   useState
 } from 'react';
 
 import { IconButton } from '../icon-button';
 import styles from './callout.module.css';
+
+/** Exit duration. Keep in sync with --rs-duration-normal (styles/effects.css:38). */
+const EXIT_MS = 200;
 
 const callout = cva(styles.callout, {
   variants: {
@@ -65,45 +69,56 @@ export function Callout({
   ...props
 }: CalloutProps) {
   // Dismissal is controlled when `onDismiss` is given; otherwise fall back to
-  // uncontrolled and hide the callout internally.
-  const [dismissed, setDismissed] = useState(false);
+  // uncontrolled: play the exit transition, then unmount.
+  const [state, setState] = useState<'open' | 'closing' | 'closed'>('open');
+
+  useEffect(() => {
+    if (state !== 'closing') return;
+    const timer = setTimeout(() => setState('closed'), EXIT_MS);
+    return () => clearTimeout(timer);
+  }, [state]);
+
   const handleDismiss = () => {
     onDismiss?.();
-    if (!onDismiss) setDismissed(true);
+    if (!onDismiss) setState('closing');
   };
-  if (dismissed) return null;
+  if (state === 'closed') return null;
 
   const role = type === 'alert' ? 'alert' : 'status';
 
   return (
-    <div
-      className={callout({ type, variant, highContrast, className })}
-      role={role}
-      aria-live={type === 'alert' ? 'assertive' : 'polite'}
-      {...props}
-    >
-      <div className={styles.container}>
-        <div className={styles.messageContainer}>
-          {icon && (
-            <div className={styles.icon} aria-hidden='true'>
-              {icon}
+    <div className={styles.transitionShell} data-state={state}>
+      <div className={styles.transitionInner}>
+        <div
+          className={callout({ type, variant, highContrast, className })}
+          role={role}
+          aria-live={type === 'alert' ? 'assertive' : 'polite'}
+          {...props}
+        >
+          <div className={styles.container}>
+            <div className={styles.messageContainer}>
+              {icon && (
+                <div className={styles.icon} aria-hidden='true'>
+                  {icon}
+                </div>
+              )}
+              <div className={styles.message}>{children}</div>
             </div>
-          )}
-          <div className={styles.message}>{children}</div>
-        </div>
 
-        <div className={styles.actionsContainer}>
-          {action && <div className={styles.action}>{action}</div>}
-          {dismissible && (
-            <IconButton
-              size={1}
-              className={styles.dismiss}
-              onClick={handleDismiss}
-              aria-label='Dismiss message'
-            >
-              <Cross1Icon />
-            </IconButton>
-          )}
+            <div className={styles.actionsContainer}>
+              {action && <div className={styles.action}>{action}</div>}
+              {dismissible && (
+                <IconButton
+                  size={1}
+                  className={styles.dismiss}
+                  onClick={handleDismiss}
+                  aria-label='Dismiss message'
+                >
+                  <Cross1Icon />
+                </IconButton>
+              )}
+            </div>
+          </div>
         </div>
       </div>
     </div>
--- a/packages/raystack/components/checkbox/checkbox.tsx
+++ b/packages/raystack/components/checkbox/checkbox.tsx
@@ -3,6 +3,7 @@
 import { Checkbox as CheckboxPrimitive } from '@base-ui/react/checkbox';
 import { CheckboxGroup as CheckboxGroupPrimitive } from '@base-ui/react/checkbox-group';
 import { cva, cx, type VariantProps } from 'class-variance-authority';
+import { useRef } from 'react';
 import { useFieldContext } from '../field';
 
 import styles from './checkbox.module.css';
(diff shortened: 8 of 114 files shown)


packages/raystack/components/headline/headline.module.css
```diff
@@ -5,7 +5,6 @@
   font-family: var(--rs-font-title);
   font-weight: var(--rs-font-weight-regular);
   color: var(--rs-color-foreground-base-primary);
-  width: 100%;
```
paanSinghCoder, line 8: **Issue:** Base `.headline` rule drops `width: 100%` here (kept only on the truncate variant), with no corresponding entry in `docs/V1-migration.md`.

**Why:** The commit itself calls this "BREAKING (mild)" for consumers centering a Headline in a flex/grid row. We've hit this exact regression before on Headline's width contract, breaking a consumer in #799 and fixed a day later in #801. A missing migration-log entry for a breaking change is also this repo's most-repeated review correction across 11 prior PRs.

**What:** Add an entry to `docs/V1-migration.md` describing this width-behavior change. Checked `not-found.tsx`, it's unaffected since it already sets `width: auto`, but other consumers relying on the old default won't get a heads up without a migration note.

packages/raystack/components/command/command.module.css
```diff
@@ -157,25 +152,17 @@
   background-color: var(--rs-color-border-base-primary);
 }
 
-.backdrop {
+.overlay {
   position: fixed;
   inset: 0;
   z-index: var(--rs-z-index-portal);
   background-color: var(--rs-color-overlay-black-a5);
-  transition:
```
paanSinghCoder, line 165: **Issue:** This `.backdrop` transition (and the matching `.dialogPopup` transition/starting-ending-style block further down) gets deleted in this PR's rename to `.overlay`, with nothing replacing it anywhere in the file.

**Why:** The commit message only claims to add a `prefers-reduced-transparency` fallback, it doesn't mention removing the base animation. Sibling `Dialog`, touched in the very same commit for the same stated purpose, kept its base opacity fade. Right now the Cmd+K palette pops open/closed instantly for everyone, not just reduced-motion users.

**What:** Restore the opacity (and scale) transition using `--rs-duration-*`/`--rs-ease-*` tokens, matching Dialog's pattern.

packages/raystack/components/data-view/components/list.tsx
```diff
@@ -346,9 +353,21 @@ export function DataViewList<TData, TValue = unknown>({
...
+        event.preventDefault(); // Space would otherwise scroll the page.
+        handleRowActivate();
+      }
+    };
     return (
       <div
-        role={ariaRole === 'table' ? 'row' : 'listitem'}
+        // A clickable row is announced as a button; a plain row keeps its
+        // structural role and stays inert for the keyboard.
+        role={onRowClick ? 'button' : ariaRole === 'table' ? 'row' : 'listitem'}
```
paanSinghCoder: **Issue:** For `variant="table"`, a clickable row now gets `role="button"` instead of `role="row"`, while its cells still render `role="cell"` and the parent stays `role="rowgroup"`.

**Why:** ARIA's required-context rule needs `cell` to have a `row` ancestor. Once the row reports as a button, most screen readers stop computing column/header association for it. Every sibling clickable-row implementation touched in this same PR (`data-table`, `data-view-beta`) kept `role="row"` for the same case.

**What:** Keep `role="row"`/`role="cell"` intact and add `tabIndex`/keydown handling alongside it, instead of swapping the role.

packages/raystack/components/field/field-root.tsx
```diff
@@ -46,15 +46,22 @@ export function FieldRoot({
...
-        {!error && description && (
-          <FieldPrimitive.Description className={styles.description}>
-            {description}
-          </FieldPrimitive.Description>
+        {(description || error) && (
+          <div className={styles.helperSlot}>
+            {description && (
+              <FieldPrimitive.Description
+                className={styles.description}
+                data-hidden={error ? '' : undefined}
```
paanSinghCoder: **Issue:** `FieldPrimitive.Description` stays mounted, just hidden via CSS, even when `error` is also set, for the new crossfade.

**Why:** Base UI registers every mounted description/error id into `aria-describedby` regardless of visibility. Verified empirically on both branches: pre-PR `aria-describedby` resolves to just the error text, post-PR to both the old hint and the error concatenated.

**What:** Stop rendering the description while an error is showing, or explicitly override `aria-describedby` to reference only the current message id.

packages/raystack/components/callout/callout.tsx
```diff
@@ -65,45 +69,56 @@ export function Callout({
...
-      role={role}
-      aria-live={type === 'alert' ? 'assertive' : 'polite'}
-      {...props}
-    >
-      <div className={styles.container}>
-        <div className={styles.messageContainer}>
-          {icon && (
-            <div className={styles.icon} aria-hidden='true'>
-              {icon}
+    <div className={styles.transitionShell} data-state={state}>
```
paanSinghCoder, line 90: **Issue:** Two new wrapper divs (`.transitionShell`, `.transitionInner`) are added around Callout's root for the exit animation, and `.transitionInner` sets `overflow: hidden` unconditionally.

**Why:** `ref` now targets the innermost div, not the actual rendered root (verified by rendering `<Callout ref={cb}>` directly). This is the same "narrowed shared component DOM shape breaks consumers" pattern that's caused two prior incidents in this repo. The unconditional `overflow: hidden` also clips overflowing content, like a focus ring, even outside the closing state.

**What:** Scope `overflow: hidden` to `[data-state="closing"]` only, and double check whether the ref-target shift is acceptable as is.

packages/raystack/components/sidebar/sidebar.module.css
```diff
@@ -82,12 +88,6 @@
   color: var(--rs-color-foreground-base-secondary);
   cursor: pointer;
   text-decoration: none;
-  transition:
```
paanSinghCoder, line 85: **Issue:** This `transition` block on `.nav-item` gets deleted with no replacement anywhere in the file.

**Why:** Every other interactive component touched in this PR was migrated onto the shared `--rs-transition-interactive` token. This one's the miss, so hover/focus/active on the sidebar's most-used control now snaps instead of easing.

**What:** Add `transition: var(--rs-transition-interactive)` back, inside the reduced-motion block like its siblings.

packages/raystack/components/sidebar/sidebar.module.css
```diff
@@ -201,13 +205,16 @@
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
-  transition: opacity 0.2s ease;
+  /* Cap the width so the collapse can animate to 0 (width: auto can't
+     transition). 200px clears any real label; the sidebar caps them narrower. */
+  max-width: 200px;
```
paanSinghCoder, line 210: **Issue:** This new `max-width` cap, and the matching transition it enables, are added so the sidebar collapse can animate, but `sidebar-item.tsx`/`sidebar-more.tsx` still conditionally unmount the `.nav-text` span in the same render the collapse fires.

**Why:** The element is gone before the transition gets a frame to run, so the label still disappears instantly. `.nav-group-label` in this same file already solves this correctly by staying mounted unconditionally.

**What:** Keep the label mounted, matching `.nav-group-label`'s pattern, instead of conditionally unmounting it.

packages/raystack/components/theme-provider/switcher.tsx
```diff
@@ -11,19 +11,29 @@ enum Theme {
...
   };
 
   return (
-    <Box {...props}>
-      {theme === Theme.DARK ? (
-        <SunIcon width={size} height={size} onClick={onClickHandler} />
+    <IconButton
+      aria-label={isDark ? 'Switch to light theme' : 'Switch to dark theme'}
+      onClick={onClickHandler}
+      style={{ width: size, height: size }}
```
paanSinghCoder, line 27: **Issue:** `IconButton`'s default padding isn't zeroed out here, while the outer element is pixel-locked to `size` via this inline style.

**Why:** The icon renders at `size - 4px`, about 13% smaller than requested by default. The existing "applies custom size" test only checks the SVG's `width` attribute, not computed layout, so it doesn't catch this.

**What:** Zero out IconButton's padding when a custom pixel `size` is supplied, or reconcile the two competing sizing mechanisms.

packages/raystack/components/checkbox/checkbox.tsx
```diff
@@ -103,11 +103,19 @@ function CheckboxItem({
...
+        keepMounted
         render={
           render ??
           ((props, state) => (
             <span {...props}>
-              {state.indeterminate ? <IndeterminateIcon /> : <CheckMarkIcon />}
+              {(state.checked ||
+                state.indeterminate ||
+                state.transitionStatus === 'ending') &&
+                (state.indeterminate ? (
```
paanSinghCoder: **Issue:** This ternary reads `state.indeterminate` from Base UI's already-updated state, even while the icon is still fading out via `transitionStatus === 'ending'`.

**Why:** When a "select all" checkbox goes indeterminate straight to unchecked, `state.indeterminate` has already flipped to false by the time the exit transition starts, so it flashes to the checkmark icon for one frame before fading, instead of fading the dash.

**What:** Track which icon was showing before the transition started, e.g. via a ref, instead of reading current state during the exit.

packages/raystack/components/tour/tour.module.css
```diff
@@ -72,7 +72,7 @@
    origin. `fade` mode repositions the card while it is hidden, so no glide. */
 @media (prefers-reduced-motion: no-preference) {
   .positioner[data-transition="move"]:has(.popup:not([data-starting-style])) {
-    transition: transform 460ms cubic-bezier(0.33, 0.7, 0, 1);
+    transition: transform var(--rs-duration-slow) var(--rs-ease-out);
```
paanSinghCoder, line 75: **Issue:** This transition maps to `--rs-ease-out`, replacing the original hand-tuned `cubic-bezier(0.33, 0.7, 0, 1)`.

**Why:** That original curve is a near-exact match for the newly-introduced `--rs-ease-drawer` token instead. The parallel rule in `tour-beta` was mapped correctly in this same PR, so this looks like a slip during consolidation rather than a deliberate change.

**What:** Use `var(--rs-ease-drawer)` instead of `var(--rs-ease-out)`.

packages/raystack/styles/typography.css
```diff
@@ -79,8 +81,6 @@
 }
 
 [data-style="modern"] {
-  --rs-font-title: var(--rs-font-inter), sans-serif;
```
paanSinghCoder, line 82: **Issue:** These two custom-property redeclarations get removed from `[data-style="modern"]`.

**Why:** A nested `<Theme style="modern">` inside an outer `<Theme style="traditional">` now inherits the outer scope's fonts instead of resetting to Inter, silently contradicting the docs' own composition example. This mirrors the `gray.css` bug this team already fixed once before (#768).

**What:** Keep the explicit `--rs-font-title`/`--rs-font-body` re-declaration here.

packages/raystack/components/table/table.tsx
```diff
@@ -25,8 +25,19 @@ function TableBody({ ...props }: ComponentProps<'tbody'>) {
 }
 TableBody.displayName = 'Table.Body';
 
-function TableRow({ ...props }: ComponentProps<'tr'>) {
-  return <tr {...props} />;
+const row = cva(styles['row'], {
+  variants: {
+    interactive: {
```
paanSinghCoder, line 30: **Issue:** This new `interactive` variant on `TableRow` isn't reflected in `apps/www/.../table/props.ts`.

**Why:** It's a real, user-facing behavioral prop, toggles hover/active row styling via CSS, and this repo's convention is to update docs `props.ts` in the same PR a component gains a new prop.

**What:** Add `interactive` to `TableRowProps` in the docs site.

packages/raystack/components/color-picker/color-picker-area.tsx
```diff
@@ -134,11 +150,64 @@ const OklchArea = ({ className, ...props }: ColorPickerAreaProps) => {
...
       onPointerDown={handlePointerDown}
+      onKeyDown={handleKeyDown}
       ref={containerRef}
+      role='slider'
+      tabIndex={0}
+      aria-label='Color area, chroma and lightness'
+      aria-valuetext={valueText}
+      aria-valuemin={0}
+      aria-valuemax={100}
+      aria-valuenow={Math.round((chroma / CHROMA_MAX) * 100)}
```
paanSinghCoder, line 210: **Issue:** `role="slider"` here reports only this one axis (chroma) via `aria-valuenow`/min/max, but the keyboard handler drives a second, independent axis (lightness) that never shows up in these values.

**Why:** `aria-valuetext` does update correctly on every keypress, so this isn't a functional break, but it's non-conformant ARIA for a genuinely 2D control.

**What:** Consider two separate slider elements, one per axis, matching established 2D color-area patterns like React Aria's ColorArea.

packages/raystack/components/calendar/date-picker.tsx
```diff
@@ -208,20 +208,29 @@ export function DatePicker({
...
+        value={inputValue}
+        onChange={handleInputChange}
+        onFocus={popover.handleInputFocus}
+        onBlur={popover.handleInputBlur}
+        onKeyUp={handleKeyUp}
+      />
+      <span
+        className={styles.datePickerError}
+        data-visible={error ? '' : undefined}
+        role='alert'
```
paanSinghCoder, line 240: **Issue:** This new error span has no `aria-describedby` link connecting it to the input.

**Why:** `role="alert"` only announces the message at the moment it appears. A user tabbing into an already-erroring field hears `aria-invalid` with no stated reason.

**What:** Give the error span a stable `id` and wire it into the input's `aria-describedby`.

packages/raystack/components/announcement-bar/announcement-bar.module.css
```diff
@@ -50,4 +50,30 @@
...
+}
+
+/* Press reads as scale, not a second dim — 0.64 text mid-press is barely
+   legible on the gradient variant, and hover's 0.8 still applies. */
+.action-btn:active {
+  transform: scale(var(--rs-scale-pressed));
+}
+
+.action-btn:focus-visible {
+  outline: 1px solid currentColor;
```
paanSinghCoder, line 67: **Issue:** This focus ring hardcodes `outline`/`currentColor` instead of the new shared `--rs-focus-ring` token used by roughly 19 other components in this PR.

**Why:** Turns out deliberate, `currentColor` reads better against the bar's colored variants per the commit message, but that reasoning lives only there, not as a code comment, so a future token-adoption pass is likely to "fix" it by mistake.

**What:** Add a one-line comment here explaining the divergence.

packages/raystack/components/avatar/avatar.tsx
```diff
@@ -164,21 +176,42 @@ const AvatarRoot = ({
...
+    if (status === 'loading') {
+      sawLoadingRef.current = true;
+      setFadeIn(false);
+    } else if (status === 'loaded') {
+      // A cached image reports 'loaded' with no 'loading' phase (Base UI never
+      // emits it), so it stays instant; only network loads fade in.
+      setFadeIn(sawLoadingRef.current);
+      sawLoadingRef.current = false;
+    }
+  };
```
rohanchkrabrty, line 192: [Base UI Avatar has `data-starting-style`](https://base-ui.com/react/components/avatar#image) css variable which can be used to fade the image. Let's use that

packages/raystack/components/announcement-bar/announcement-bar.module.css
```diff
@@ -50,4 +50,30 @@
...
+  outline-offset: 2px;
+}
```
rohanchkrabrty, line 69: Right now the action-btn is a html button. Ideally it should be using Apsara Button component, in that way we don't have to add focus styles in other components.

packages/raystack/styles/effects.css
```diff
@@ -1,35 +1,101 @@
...
+    0.22,
+    1,
+    0.36,
+    1
+  ); /* entrances, exits, value changes */
+  --rs-ease-in-out: cubic-bezier(
+    0.77,
+    0,
+    0.175,
+    1
```
rohanchkrabrty: lines need to be prettified/formatted

packages/raystack/components/breadcrumb/breadcrumb.module.css
```diff
@@ -39,10 +39,23 @@
...
   color: var(--rs-color-foreground-base-secondary);
 }
 
+.breadcrumb-link:focus-visible {
+  outline: none;
+  background-color: var(--rs-color-background-base-primary-hover);
+  color: var(--rs-color-foreground-base-secondary);
+  border-radius: var(--rs-radius-1);
+}
+
```
rohanchkrabrty, line 56: the breadcrumb focus visible looks weird and not consistent. Let's make it consistent with other styles. Something similar to - 

Image

packages/raystack/components/breadcrumb/breadcrumb.module.css
```diff
@@ -81,6 +94,12 @@
   padding: 0;
 }
 
+.breadcrumb-dropdown-trigger:focus-visible {
+  background-color: var(--rs-color-background-base-primary-hover);
+  color: var(--rs-color-foreground-base-primary);
+  border-radius: var(--rs-radius-1);
+}
+
```
rohanchkrabrty, line 99: Same for this

packages/raystack/components/accordion/accordion.module.css
```diff
@@ -31,11 +31,20 @@
...
-.accordion-trigger:hover,
-.accordion-trigger:focus-visible {
+.accordion-trigger:hover {
   background-color: var(--rs-color-background-base-primary-hover);
 }
 
+/* Trigger sits flush in a clipped container: keep the ring inside. */
+.accordion-trigger:focus-visible {
+  outline: var(--rs-focus-ring);
+  outline-offset: var(--rs-focus-ring-offset-inset);
```
rohanchkrabrty: The offset needs to be `-1px` in this case, else the border and outline both are visible.

with -2px :

Image


with 1px:

Image


Let's create two variables for the `outline-inset` , supporting both 1px & 2px

packages/raystack/styles/effects.css
```diff
@@ -1,35 +1,101 @@
...
-    --rs-blur-md: blur(1px);
-    --rs-blur-lg: blur(2px);
-    --rs-blur-xl: blur(4px);
-}
+  /* Focus ring: keyboard-focus indicator. The shadow variant is for
+       chromeless text-entry surfaces that have no border to ring. Accent by
+       default; a colored control (danger, success, …) overrides
+       `outline-color` in its own :focus-visible rule to match. */
+  --rs-focus-ring: 1px solid var(--rs-color-border-accent-emphasis);
+  --rs-focus-ring-shadow: 0 0 0 1.5px var(--rs-color-border-accent-primary);
```
rohanchkrabrty:
```suggestion
  --rs-focus-ring-shadow: 0 0 0 1px var(--rs-color-border-accent-emphasis);
```

^ This is used mainly for borderless input variants. We should match this style to the ring css for consistency.

rohanchkrabrty (changes requested):

packages/raystack/components/announcement-bar/announcement-bar.module.css
ravisuhag: Tried swapping in `<Button variant="text">`. It doesn't fit cleanly here:

- Button's `text` variant hardcodes `color: var(--rs-color-foreground-base-primary)` (not `currentColor`), so on the normal bar (dark bg) the label goes dark-on-dark and on the error bar it's low contrast. To keep it readable I had to override color, background, and the hover box — so most of the custom CSS stayed anyway.
- The one thing we'd inherit, the focus ring, is `1px solid --rs-color-border-accent-emphasis` (blue). The current ring is `currentColor`, so it matches each bar (white on dark/error, dark on gradient). A blue ring on the red error bar looks off-brand and lower contrast — so we'd want to re-add a currentColor override regardless.

Net: the swap adds a Button dependency without removing the custom styles, and slightly regresses the focus ring on colored bars. Keeping the raw button for now. In long term either we can use scoped theming to revert or currentColor option.

packages/raystack/components/avatar/avatar.tsx
ravisuhag: Looked into this. `data-starting-style` fades **every** image on load — it can't tell a cached image from a network one. So it would make cached avatars fade in too, which flickers in lists where avatars mount and unmount a lot (like tables).

The current code keeps cached avatars instant and only fades real network loads, which is the behaviour we want. Keeping it as-is.

packages/raystack/components/headline/headline.module.css
ravisuhag: Added a migration note under the Headline section of V1-migration.md covering the base width:100% removal and the one-line opt-back-in. Done.

packages/raystack/components/command/command.module.css
ravisuhag: The backdrop was removed from the palette entirely, so the overlay fade no longer applies. Restored the popup open/close fade + scale (tokenized, reduced-motion gated) so it no longer pops open instantly. Done.

packages/raystack/components/data-view/components/list.tsx
ravisuhag: Kept clickable rows as role=row/listitem so cells stay associated with their row, with tabIndex + keydown for activation. Matches data-view-beta. Done.

packages/raystack/components/field/field-root.tsx
ravisuhag: The description is now unmounted while an error is showing, so aria-describedby resolves to just the error. Done.

packages/raystack/components/callout/callout.tsx
ravisuhag: Scoped overflow:hidden to [data-state=closing] so open content (focus rings, popovers) is no longer clipped. Left the ref target as-is: it points to the same styled callout element it did before this PR, so consumers are unaffected. Done.

packages/raystack/components/sidebar/sidebar.module.css
ravisuhag: Leaving this as-is. The instant nav-item hover/press is intentional to avoid a laggy feel on the sidebar most-used control.

packages/raystack/components/sidebar/sidebar.module.css
ravisuhag: Kept the label mounted (matching nav-group-label) so it collapses with the sidebar instead of popping out. Done.

packages/raystack/components/theme-provider/switcher.tsx
ravisuhag: Reconciled by treating size as the button box and letting IconButton CSS size the icon (dropped the redundant width/height on the icons). Updated the test to assert the button box. Done.

packages/raystack/components/checkbox/checkbox.tsx
ravisuhag: Now tracks the shown icon in a ref and renders it during the exit, so indeterminate to unchecked fades the dash instead of flashing the check for a frame. Done.

packages/raystack/components/tour/tour.module.css
ravisuhag: Skipping. tour-beta parallel rule also uses --rs-ease-out (not ease-drawer), so changing only tour would make the two inconsistent. Leaving both on ease-out for now.

packages/raystack/styles/typography.css
ravisuhag: Restored the --rs-font-title/--rs-font-body re-declaration on [data-style=modern] so a nested modern theme resets to Inter instead of inheriting the outer fonts. Done.

packages/raystack/components/table/table.tsx
ravisuhag: Added interactive to TableRowProps in the docs props.ts. Done.

packages/raystack/components/color-picker/color-picker-area.tsx
ravisuhag: Skipping for this PR. aria-valuetext already announces both axes so it is not a functional break, and the two-slider refactor (which also affects HslArea) is better as its own change. Noting as a follow-up.

packages/raystack/components/calendar/date-picker.tsx
ravisuhag: Gave the error span a stable id and wired it into the input aria-describedby (merged with any consumer value) when an error is present. Done.

packages/raystack/components/accordion/accordion.module.css
ravisuhag: Fixed. Added a --rs-focus-ring-offset-inset-border (-1px) token and pointed accordion at it so the ring hugs the border. Applied the same to number-field steppers and the grouped toggle (same bordered case). Done.

packages/raystack/components/breadcrumb/breadcrumb.module.css
ravisuhag: Fixed. Swapped the background-fill focus for the shared --rs-focus-ring outline, matching Link. Done.

packages/raystack/components/breadcrumb/breadcrumb.module.css
ravisuhag: Same fix applied to the dropdown trigger focus. Done.

packages/raystack/styles/effects.css
ravisuhag: Fixed. Moved the trailing comments above each easing token so the values stay on single lines and Biome keeps them that way. Done.

packages/raystack/components/data-view/components/list.tsx
```diff
@@ -346,9 +353,21 @@ export function DataViewList<TData, TValue = unknown>({
...
       ? { ...style, gridTemplateColumns }
       : (style ?? {});
+    const handleRowActivate = () => onRowClick?.(row.original);
+    const handleRowKeyDown = (event: KeyboardEvent<HTMLDivElement>) => {
+      // Enter and Space both activate, matching a native button.
+      if (event.key === 'Enter' || event.key === ' ') {
+        event.preventDefault(); // Space would otherwise scroll the page.
+        handleRowActivate();
+      }
+    };
```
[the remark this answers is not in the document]
ravisuhag: Fixed. Added a target===currentTarget guard so keydowns bubbling from interactive children in a row do not also trigger row activation. Done.

packages/raystack/components/avatar/avatar.tsx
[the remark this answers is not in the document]
ravisuhag: Skipping. Valid edge case but low-frequency (image error, then src change to a cached image); leaving the current fade logic as-is for this PR.

packages/raystack/components/copy-button/copy-button.module.css
```diff
@@ -0,0 +1,57 @@
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
```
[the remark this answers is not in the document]
ravisuhag: Skipping. clip: rect(...) still works across all target browsers; not switching to clip-path in this PR.

packages/raystack/components/switch/switch.module.css
```diff
@@ -73,3 +83,20 @@
...
+}
+
+.switch[data-checked]:active:not([data-disabled]) .thumb {
+  transform: translateX(var(--rs-space-5)) scaleX(1.15);
+  transform-origin: right center;
+}
+
+.switch.small[data-checked]:active:not([data-disabled]) .thumb {
+  transform: translateX(var(--rs-space-4)) scaleX(1.15);
+}
```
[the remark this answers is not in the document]
ravisuhag: Not a bug. transform-origin: right center already applies to the small checked pressed state via the less-specific .switch[data-checked]:active rule, so it is not missing.

packages/raystack/components/announcement-bar/announcement-bar.module.css
```diff
@@ -50,4 +50,30 @@
...
+
+.action-btn:hover {
+  opacity: 0.8;
+}
+
+/* Press reads as scale, not a second dim — 0.64 text mid-press is barely
+   legible on the gradient variant, and hover's 0.8 still applies. */
+.action-btn:active {
+  transform: scale(var(--rs-scale-pressed));
+}
```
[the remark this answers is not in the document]
ravisuhag: Skipping. The press scale has no transition under reduced-motion (it applies instantly), and the announcement-bar styling is staying as-is.

packages/raystack/components/announcement-bar/announcement-bar.module.css
```diff
@@ -50,4 +50,19 @@
...
+  opacity: 0.8;
+}
+
+.action-btn:active {
+  opacity: 0.64;
+}
+
+.action-btn:focus-visible {
+  outline: 1px solid currentColor;
+  outline-offset: 2px;
```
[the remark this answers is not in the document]
ravisuhag: Skipping. currentColor is deliberate so the ring matches the bar colored variants. The repo formats with Biome (no stylelint), which is happy with the casing.

CONTRIBUTING.md
```diff
@@ -222,10 +222,10 @@ Apsara follows an automated release process using GitHub Actions and semantic ve
    # Ensure you're on the correct branch
    git checkout main  # for production release
    # OR
-   git checkout develop  # for release candidate
+   git checkout release/x.y  # for release candidate
    
    # Pull latest changes
-   git pull origin main  # or develop
+   git pull origin main  # or the release branch
```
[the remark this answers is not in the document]
ravisuhag: Skipping. Minor docs wording, out of scope for this design-system PR.

packages/raystack/components/command/command.module.css
```diff
@@ -202,13 +194,12 @@
...
-.dialogPopup[data-starting-style],
-.dialogPopup[data-ending-style] {
-  opacity: 0;
-  transform: scale(0.98);
+/* Reduced transparency: drop the blur, compensate with a heavier solid scrim */
+@media (prefers-reduced-transparency: reduce) {
+  .overlayBlur {
+    backdrop-filter: none;
+    background-color: var(--rs-color-overlay-black-a9);
+  }
```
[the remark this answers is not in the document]
ravisuhag: Skipping. No stylelint in this repo (Biome is the formatter and reports no issue here); the command styles were also reworked in this PR.

packages/raystack/components/drawer/drawer.module.css
```diff
@@ -18,7 +18,10 @@
 }
 
 .backdrop[data-ending-style] {
-  transition-duration: calc(var(--drawer-swipe-strength, 1) * 400ms);
+  transition-duration: calc(
+    var(--drawer-swipe-strength, 1) *
+    var(--rs-duration-drawer)
+  );
```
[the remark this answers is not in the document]
ravisuhag: Skipping. This stylelint rule is not used in the repo; Biome is the formatter and is clean here.

packages/raystack/components/link/link.module.css
```diff
@@ -1,9 +1,39 @@
...
+  .link:active {
+    transition: opacity var(--rs-duration-press) var(--rs-ease-out);
+  }
 }
 
 .link:hover {
-  opacity: 0.8;
+  text-decoration-color: currentColor;
+}
+
```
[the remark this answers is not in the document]
ravisuhag: Skipping. No stylelint in the repo; Biome (the formatter) is happy with the casing.

packages/raystack/components/meter/meter.module.css
```diff
@@ -53,7 +55,10 @@
 
 .circularSvg {
   --rs-meter-track-size: var(--rs-space-2);
-  --rs-meter-radius: calc((var(--rs-space-14) - var(--rs-meter-track-size) * 2) / 2);
+  --rs-meter-radius: calc(
+    (var(--rs-space-14) - var(--rs-meter-track-size) * 2) /
+    2
+  );
```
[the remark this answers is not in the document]
ravisuhag: Skipping. No stylelint in the repo; Biome is clean here.

packages/raystack/components/progress/progress.module.css
```diff
@@ -53,7 +55,10 @@
 
 .circularSvg {
   --rs-progress-track-size: var(--rs-space-2);
-  --rs-progress-radius: calc((var(--rs-space-14) - var(--rs-progress-track-size) * 2) / 2);
+  --rs-progress-radius: calc(
+    (var(--rs-space-14) - var(--rs-progress-track-size) * 2) /
+    2
+  );
```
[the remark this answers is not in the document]
ravisuhag: Skipping. No stylelint in the repo; Biome is clean here.

packages/raystack/components/select/select-trigger.tsx
```diff
@@ -4,7 +4,7 @@ import {
   Combobox as ComboboxPrimitive,
   Select as SelectPrimitive
 } from '@base-ui/react';
-import { ChevronDownIcon } from '@radix-ui/react-icons';
+import { CaretSortIcon } from '@radix-ui/react-icons';
```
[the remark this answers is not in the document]
ravisuhag: The trigger now imports ChevronDownIcon rather than CaretSortIcon, so this mock note is out of date. Tests pass. Skipping.

packages/raystack/components/toast/toast.module.css
```diff
@@ -66,7 +66,7 @@
   cursor: default;
   z-index: calc(1000 - var(--toast-index));
 
-  transition: opacity 500ms;
+  transition: opacity var(--rs-duration-slow) var(--rs-ease-out);
```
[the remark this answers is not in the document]
ravisuhag: Skipping. No stylelint in the repo; Biome is clean here.

packages/raystack/styles/effects.css
ravisuhag: Applied your suggestion. --rs-focus-ring-shadow is now 0 0 0 1px var(--rs-color-border-accent-emphasis), matching the outline ring. Affects the three borderless inputs (input, text-area, otp-field). Done in 57336d5b.

packages/raystack/components/announcement-bar/announcement-bar.module.css
ravisuhag: Skipping. The currentColor focus ring here is deliberate so it reads against the bar colored variants; leaving it as-is for this PR.

rohanchkrabrty (approved):

PR #860, opened 2026-07-21
feat: dataview timeline component

rohanchkrabrty:
## Summary

- Adds `DataView.Timeline`, a renderer that positions the shared DataView row model as cards on a continuous, horizontally scrollable time axis — the card interior is fully consumer-owned via `renderCard(row, context)`, while the Timeline owns the time scale, lane packing, sticky two-tier axis (ticks, month/year bands, today + custom markers, gridlines), and native x/y scrolling with drag-to-pan, momentum glide, and a snapped hover cursor.
- Supports range-window data loading: explicit `range` for a stable coordinate space, `onVisibleRangeChange` (rAF-throttled, deduped by ms window), and time-anchored scrolling so the content never jumps when the domain shifts; imperative navigation ships via `actionsRef` (`scrollTo`, `getVisibleRange`).
- Performance: memoized card wrapper so per-frame cursor/viewport state never re-renders card interiors, and binary-search horizontal culling behind `virtualized`.
- Fixes from review: scroll position is restored when a gated view is re-activated, and point-marker lanes pack with `estimatedPointWidth` so content-sized cards don't overlap; `rowHeight` is renamed to `estimatedRowHeight` to match `DataView.List`.
- Includes docs (guide + props tables), playground demos, an order-desk example page, and 52 unit tests covering geometry, axis, panning, culling, and the imperative handle.

### DIFF
--- a/apps/www/src/app/examples/timeline/page.tsx
+++ b/apps/www/src/app/examples/timeline/page.tsx
@@ -84,6 +84,12 @@ const NAME_POOL = [
   'Order_9'
 ];
 
+/* Orders generated per month. Deliberately dense: enough cards for lane
+   packing to stack several deep inside every status band, which is what makes
+   the grouped swim-lane layout (and horizontal culling) worth looking at. */
+const ORDERS_PER_MONTH_MIN = 18;
+const ORDERS_PER_MONTH_MAX = 26;
+
 const STATUS_LABEL: Record<OrderStatus, string> = {
   scheduled: 'Scheduled',
   'in-progress': 'In progress',
@@ -116,7 +122,9 @@ function ordersForMonth(monthKey: string): Order[] {
   const monthStart = dayjs(`${monthKey}-01`);
   const rng = mulberry32(monthStart.year() * 100 + monthStart.month());
   const today = startOfToday();
-  const count = 5 + Math.floor(rng() * 4); // 5–8 orders per month
+  const count =
+    ORDERS_PER_MONTH_MIN +
+    Math.floor(rng() * (ORDERS_PER_MONTH_MAX - ORDERS_PER_MONTH_MIN + 1));
   return Array.from({ length: count }, (_, i) => {
     const start = monthStart.add(Math.floor(rng() * 27), 'day');
     const due = start.add(2 + Math.floor(rng() * 16), 'day');
@@ -187,6 +195,11 @@ const fields: DataViewField<Order>[] = [
     filterable: true,
     filterType: 'select',
     hideable: true,
+    // Groupable fields drive the Grouping control in Display settings. The
+    // timeline renders one swim-lane section per group; the list view renders
+    // the same groups as section headers.
+    groupable: true,
+    showGroupCount: true,
     filterOptions: ORG_LIST.map(org => ({ value: org, label: org }))
   },
   {
@@ -195,6 +208,10 @@ const fields: DataViewField<Order>[] = [
     filterable: true,
     filterType: 'select',
     hideable: true,
+    groupable: true,
+    showGroupCount: true,
+    // Bucket keys are the raw status values; the band shows these labels.
+    groupLabelsMap: STATUS_LABEL,
     filterOptions: (Object.keys(STATUS_LABEL) as OrderStatus[]).map(status => ({
       value: status,
       label: STATUS_LABEL[status]
@@ -707,6 +724,10 @@ const Page = () => {
           isLoading={isLoading}
           loadingRowCount={3}
           defaultSort={{ name: 'dueDate', order: 'asc' }}
+          /* Opens grouped by status — one swim-lane band per status in the
+             timeline, the same buckets as the list view's group headers.
+             Switch it in Display settings → Grouping. */
+          query={{ group_by: ['status'] }}
           getRowId={(row: Order) => row.id}
           onRowClick={order => orderDialog.openWithPayload(order)}
           views={[
@@ -770,7 +791,12 @@ const Page = () => {
                 <OrderCard order={row.original} context={context} />
               )}
             />
-            <DataView.List name='list' variant='table' columns={orderColumns} />
+            <DataView.List
+              name='list'
+              variant='table'
+              columns={orderColumns}
+              stickyGroupHeader
+            />
             <DataView.EmptyState>
               <EmptyState
                 icon={<FilterIcon />}
--- a/apps/www/src/app/examples/timeline/page.tsx
+++ b/apps/www/src/app/examples/timeline/page.tsx
@@ -1,0 +1,788 @@
+/** biome-ignore-all lint/suspicious/noShadowRestrictedNames: public component name intentionally matches the package export */
+'use client';
+
+import {
+  Badge,
+  Button,
+  DataView,
+  type DataViewField,
+  type DataViewListColumn,
+  Dialog,
+  EmptyState,
+  Flex,
+  IconButton,
+  Navbar,
+  Sidebar,
+  Text,
+  type TimelineActions,
+  type TimelineCardContext
+} from '@raystack/apsara';
+import { BellIcon, FilterIcon, SidebarIcon } from '@raystack/apsara/icons';
+import dayjs from 'dayjs';
+import {
+  type ReactNode,
+  useCallback,
+  useEffect,
+  useRef,
+  useState
+} from 'react';
+
+type OrderStatus = 'scheduled' | 'in-progress' | 'partial' | 'completed';
+
+type Order = {
+  id: string;
+  name: string;
+  org: string;
+  priority: number;
+  status: OrderStatus;
+  startDate: string;
+  dueDate: string;
+};
+
+/* Dates are generated relative to today so the today-line and the
+   "Due in N days" danger states are always live in the example. Pinned to
+   midnight so server- and client-rendered output match (no hydration drift). */
+const DAY_MS = 86_400_000;
+const startOfToday = () => {
+  const now = new Date();
+  now.setHours(0, 0, 0, 0);
+  return now.getTime();
+};
+const daysFromNow = (days: number) =>
+  new Date(startOfToday() + days * DAY_MS).toISOString();
+
+/* ── Programmatic scroll domain (min/max of the timeline) ─────────────────
+   The explorable window is fixed to ±6 months around today. An explicit
+   `range` keeps the coordinate space stable while data streams in — the
+   scrollbar never jumps as new rows load. */
+const RANGE: [string, string] = [daysFromNow(-183), daysFromNow(183)];
+
+/* How far beyond the visible window to prefetch on each scroll event. */
+const PREFETCH_MS = 14 * DAY_MS;
+/* Minimum spacing between range fetches while scrolling. */
+const FETCH_THROTTLE_MS = 200;
+/* Simulated network latency for the fake orders API. */
+const API_LATENCY_MS = 500;
+
+const ORG_LIST = [
+  'Company A',
+  'Company B',
+  'Company C',
+  'Company D',
+  'Company E'
+];
+
+const NAME_POOL = [
+  'Order_1',
+  'Order_2',
+  'Order_3',
+  'Order_4',
+  'Order_5',
+  'Order_6',
+  'Order_7',
+  'Order_8',
+  'Order_9'
+];
+
+const STATUS_LABEL: Record<OrderStatus, string> = {
+  scheduled: 'Scheduled',
+  'in-progress': 'In progress',
+  partial: 'Partially delivered',
+  completed: 'Completed'
+};
+
+/* ── Simulated orders backend ──────────────────────────────────────────────
+   The full dataset is generated once up front (the "database table"), then
+   `fetchOrdersApi` answers range queries against it with the interval-overlap
+   predicate — the same shape a real RQL backend would run:
+
+     start_time <= to AND end_time >= from
+
+   i.e. a card is returned iff it has at least one visible pixel in the
+   window. A containment filter (`start >= from AND end <= to`) would drop
+   cards that straddle either window edge. */
+function mulberry32(seed: number) {
+  let a = seed;
+  return () => {
+    a |= 0;
+    a = (a + 0x6d2b79f5) | 0;
+    let t = Math.imul(a ^ (a >>> 15), 1 | a);
+    t = (t + Math.imul(t ^ (t >>> 7), 61 | t)) ^ t;
+    return ((t ^ (t >>> 14)) >>> 0) / 4294967296;
+  };
+}
+
+function ordersForMonth(monthKey: string): Order[] {
+  const monthStart = dayjs(`${monthKey}-01`);
+  const rng = mulberry32(monthStart.year() * 100 + monthStart.month());
+  const today = startOfToday();
+  const count = 5 + Math.floor(rng() * 4); // 5–8 orders per month
+  return Array.from({ length: count }, (_, i) => {
+    const start = monthStart.add(Math.floor(rng() * 27), 'day');
+    const due = start.add(2 + Math.floor(rng() * 16), 'day');
+    let status: OrderStatus;
+    if (due.valueOf() < today) {
+      status = rng() < 0.75 ? 'completed' : 'partial';
+    } else if (start.valueOf() > today) {
+      status = 'scheduled';
+    } else {
+      status = rng() < 0.5 ? 'in-progress' : 'partial';
+    }
+    return {
+      id: `${monthKey}-${i}`,
+      name: NAME_POOL[Math.floor(rng() * NAME_POOL.length)],
+      org: ORG_LIST[Math.floor(rng() * ORG_LIST.length)],
+      priority: 1 + Math.floor(rng() * 10),
+      status,
+      startDate: start.toISOString(),
+      dueDate: due.toISOString()
+    };
+  });
+}
+
+function monthKeysBetween(fromMs: number, toMs: number): string[] {
+  const keys: string[] = [];
+  let cursor = dayjs(fromMs).startOf('month');
+  const end = dayjs(toMs);
+  while (cursor.isBefore(end)) {
+    keys.push(cursor.format('YYYY-MM'));
+    cursor = cursor.add(1, 'month');
+  }
+  return keys;
+}
+
+/* The whole "table", generated once at module load. Seeded per month, so the
+   data is stable across renders and reloads within a day. */
+const ALL_ORDERS: Order[] = monthKeysBetween(
+  Date.parse(RANGE[0]),
+  Date.parse(RANGE[1])
+).flatMap(ordersForMonth);
+
+function fetchOrdersApi(fromMs: number, toMs: number): Promise<Order[]> {
+  // Interval-overlap predicate: starts before the window closes AND ends
+  // after it opens — catches fully-inside cards and edge-spanners alike.
+  const rows = ALL_ORDERS.filter(
+    order =>
+      Date.parse(order.startDate) <= toMs && Date.parse(order.dueDate) >= fromMs
+  ).sort((a, b) => Date.parse(a.startDate) - Date.parse(b.startDate));
+  return new Promise(resolve =>
+    setTimeout(() => resolve(rows), API_LATENCY_MS)
+  );
+}
+
+// Renderer-agnostic metadata — drives search, filters, and the Display
+// Properties toggles that the card composes via <DataView.DisplayAccess>.
+const fields: DataViewField<Order>[] = [
+  {
+    accessorKey: 'name',
+    label: 'Order',
+    filterable: true,
+    filterType: 'string',
+    sortable: true,
+    hideable: false
+  },
+  {
+    accessorKey: 'org',
+    label: 'Organization',
+    filterable: true,
+    filterType: 'select',
+    hideable: true,
+    filterOptions: ORG_LIST.map(org => ({ value: org, label: org }))
+  },
+  {
+    accessorKey: 'status',
+    label: 'Status',
+    filterable: true,
+    filterType: 'select',
+    hideable: true,
+    filterOptions: (Object.keys(STATUS_LABEL) as OrderStatus[]).map(status => ({
+      value: status,
+      label: STATUS_LABEL[status]
+    }))
+  },
+  { accessorKey: 'priority', label: 'Priority', hideable: true },
+  {
+    accessorKey: 'startDate',
+    label: 'Start date',
+    filterable: true,
+    filterType: 'date',
+    sortable: true,
+    hideable: true
+  },
+  {
+    accessorKey: 'dueDate',
+    label: 'Due date',
+    filterable: true,
+    filterType: 'date',
+    sortable: true,
+    hideable: true
+  }
+];
+
+/* Status icons matching the Figma card anatomy: outlined circle (scheduled),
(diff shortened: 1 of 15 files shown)


packages/raystack/components/data-view/utils/time-scale.tsx
```diff
@@ -0,0 +1,220 @@
...
+}): TimelineTimeScale {
+  const {
+    minTime,
+    maxTime,
+    scale,
+    unitWidth,
+    padUnits = 0,
+    minWidth = 0
+  } = params;
+  const pxPerMs = unitWidth / TIMELINE_UNIT_MS[scale];
```
rohilsurana, line 91: **[robustness]** `unitWidth` is consumer-controllable (the `unitWidth` prop). If it is `0` or negative, `pxPerMs` becomes `0`, and the viewport-fill step below computes `Math.ceil(deficitPx / unitWidth)` = `Infinity`, which feeds `addUnits(end, scale, Infinity)` and a `while` loop that can never exit. A single bad prop value hangs the render. Suggest clamping at the boundary, e.g. `const unitWidth = Math.max(1, params.unitWidth)`. This is the one point I would treat as a real gate, because the failure mode is a hang rather than a wrong pixel.

packages/raystack/components/data-view/utils/time-scale.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const effectiveLabelEvery = Math.max(
+    autoLabelEvery,
+    Math.floor(labelEvery ?? 1)
+  );
+
+  const ticks: TimelineTick[] = [];
+  let cursor = startOfUnit(dayjs(timeScale.t0), scale);
+  if (cursor.valueOf() < timeScale.t0) cursor = addUnits(cursor, scale, 1);
+  let index = 0;
+  while (cursor.valueOf() <= timeScale.t1) {
```
rohilsurana, line 187: **[scalability, non-blocking]** `buildAxis` materializes one tick object per unit across the whole domain. The `virtualized` path culls what *renders* but not what gets *built* here, so a `day` scale over a multi-year `range` allocates tens of thousands of ticks on every domain rebuild. Fine for the shipped use cases (weeks/months). Worth a code comment so nobody points it at a decade of days and wonders why it stalls.

packages/raystack/components/data-view/utils/pack-lanes.tsx
```diff
@@ -0,0 +1,43 @@
...
+  width: number;
+}
+
+export interface PackLanesResult {
+  /** Lane index per input item, in the input's original order. */
+  lanes: number[];
+  laneCount: number;
+}
+
+const DEFAULT_LANE_GAP_PX = 8;
```
rohilsurana: **[minor]** Two things: (1) the horizontal in-lane gap defaults to 8px and the component never passes a value, so it is not configurable, while the vertical gap (`laneGap`) is a public prop defaulting to 16. Worth confirming that asymmetry is intentional. (2) `DEFAULT_LANE_GAP_PX` (8, horizontal) reads almost identically to `DEFAULT_LANE_GAP` (16, vertical) in timeline.tsx but they are different axes; a rename like `DEFAULT_CARD_GAP_PX` would remove the trap for the next reader. For the record, the first-fit greedy packing itself is correct and optimal in lane count.

packages/raystack/components/data-view/data-view.types.tsx
```diff
@@ -202,6 +203,160 @@ export interface DataViewListProps<TData, TValue = unknown> {
...
+  /** Receives the imperative navigation handle (`scrollTo`, `getVisibleRange`). */
+  actionsRef?: React.RefObject<TimelineActions | null>;
+
+  /**
+   * 'auto' (default) packs non-overlapping cards into shared lanes (greedy
+   * interval scheduling); 'one-per-row' gives every row its own lane.
+   */
+  lanePacking?: 'auto' | 'one-per-row';
+  /** Card height in px (cards render at exactly this height; named to match `DataView.List`). Default 66. */
+  estimatedRowHeight?: number;
```
rohilsurana, line 379: **[api / naming]** `rowHeight` was renamed to `estimatedRowHeight` to match `DataView.List`, but the semantics differ. In `List`, `estimatedRowHeight` is a true estimate: rows auto-measure after paint and it is only used until the first measurement. In `Timeline`, cards render at exactly this height (applied as a hard `height` style; nothing measures), so content taller than this value clips or overflows. The shared name promises auto-measurement that does not happen here. Options: keep `rowHeight` (honest), or keep the name but make the doc explicit that cards do not auto-measure unlike `List`. Since the rename came from earlier review feedback, flagging the mismatch rather than deciding it.

packages/raystack/components/data-view/data-view.types.tsx
```diff
@@ -202,6 +203,160 @@ export interface DataViewListProps<TData, TValue = unknown> {
...
+  /** Accessor key for the end date. Omitted → point markers; present → variable-width span cards. */
+  endField?: string;
+
+  /**
+   * Renders the card interior. The Timeline owns positioning (x from start,
+   * width from span, lane from packing, scroll); the consumer owns the card
+   * visual entirely — chrome, states, truncation, and the collapsed variant.
+   * Compose `<DataView.DisplayAccess>` inside for Display Properties support.
+   */
+  renderCard: (
```
rohilsurana, line 312: **[perf / docs]** The Timeline memoizes its card wrapper (`TimelineCardView`) so per-frame cursor/viewport state does not re-render card interiors. That optimization only holds if `renderCard` is referentially stable. If a consumer passes an inline `renderCard={(row, ctx) => ...}`, its identity changes every render, the memo never hits, and every visible card re-renders on every scroll frame. Nothing here tells consumers to wrap it in `useCallback`. Suggest adding that guidance to this doc (the same applies to `onRowClick`).

packages/raystack/components/data-view/data-view.types.tsx
```diff
@@ -202,6 +203,160 @@ export interface DataViewListProps<TData, TValue = unknown> {
...
+/**
+ * Geometry + state handed to `renderCard`. The Timeline owns positioning; the
+ * consumer owns the card visual and uses this context to adapt it (e.g. render
+ * a compact stub when `collapsed`).
+ */
+export interface TimelineCardContext {
+  /** Pixel width of the time span (0 when `endField` is omitted). */
+  width: number;
+  /** True when the span is narrower than `minCardWidth`. */
+  collapsed: boolean;
```
rohilsurana, line 232: **[minor doc]** `collapsed` is documented as true when the span is narrower than `minCardWidth`, but point cards (no `endField`) are never collapsed regardless of width (see `collapsed={item.endTime !== null && ...}` in timeline.tsx). One clause here would close the gap.

packages/raystack/components/data-view/components/timeline.tsx
```diff
@@ -0,0 +1,1061 @@
...
+      style={style}
+      data-collapsed={collapsed || undefined}
+      onClick={onRowClick ? () => onRowClick(row.original) : undefined}
+    >
+      {renderCard(row, context)}
+    </div>
+  );
+}
+
+const TimelineCardView = memo(
```
rohilsurana, line 217: **[perf]** The value of this `memo` depends on `row`, `renderCard`, and `onRowClick` being referentially stable across renders (the comment above says as much). `row` is stable from TanStack, but `renderCard`/`onRowClick` come straight from consumer props with no memoization enforced. If a consumer passes them inline, the shallow compare fails every frame and this memo becomes pure overhead. Pairs with the `renderCard` doc note in data-view.types.tsx. No change needed in this file; calling it out so the consumer-facing docs cover it.

packages/raystack/components/data-view/components/timeline.tsx
```diff
@@ -0,0 +1,1061 @@
...
+      )
+    : ticks;
+  const cardClassName = cx(
+    styles.timelineCard,
+    onRowClick && styles.clickable,
+    classNames.card
+  );
+
+  return (
+    <div
```
rohilsurana, line 1196: **[a11y]** The scroll container is the primary way to navigate the timeline, but it is not keyboard operable: no `tabIndex`, so keyboard-only users cannot focus it to scroll horizontally with arrow keys, and drag-to-pan is mouse-only by design. The `role="list"` canvas also has no accessible name. Focusable content inside `renderCard` lets tab reach individual cards, but panning the axis itself is unreachable without a pointer. Consider `tabIndex={0}` + an `aria-label` on the scroll region, or documenting the keyboard story. Medium priority.

packages/raystack/components/data-view/components/timeline.tsx
```diff
@@ -0,0 +1,1061 @@
...
+      // Throws for pointers the browser isn't tracking (synthetic events).
+      try {
+        el.setPointerCapture?.(event.pointerId);
+      } catch {
+        /* noop */
+      }
+      // Panning, not selecting — suppress native text-selection drag.
+      event.preventDefault();
+    },
+    []
```
rohilsurana: **[nit]** `handleDragPointerDown` uses `stopMomentum` but its dependency array is `[]`. It works because `stopMomentum` is a stable `useCallback([])`, but it is inconsistent with the rest of this file, which is otherwise careful with exhaustive deps (and would trip `react-hooks/exhaustive-deps` if that rule is on). Either add `stopMomentum` to the array or leave a note.

rohilsurana (commented):
## Review: `DataView.Timeline` (in progress)

Reviewing this outside-in over a few passes. Covered so far: the public API/types, the pure utils (`time-scale`, `pack-lanes`), and the component itself. Styles, the 52 tests, and the docs/demo are still to come.

Overall this is strong, careful work. The renderer contract mirrors `DataView.List` and the `actionsRef` pattern from `tour-beta` closely, so the API stays consistent with the rest of the library. The scroll/pan/momentum machinery and the scroll-anchoring on domain shift are thoughtfully done, hooks are all called before the early returns, and rAF throttling plus cleanup are handled properly.

Confirmed the data scrub this PR exists for: the only change from the previous PR is the demo mock data in `examples/timeline/page.tsx`, real org/file names are gone, and no proprietary terms remain in any of the 15 files or in the branch's own commits. Note this only protects `main` going forward; the earlier data still lives in the previous PR's branch and commits on the remote if that matters.

Inline notes below, tagged by severity. The only one I would treat as a real gate is the `unitWidth <= 0` guard in `time-scale.tsx`. The rest are API/doc/a11y polish. Nothing here blocks the concept.

packages/raystack/components/data-view/data-view.module.css
```diff
@@ -344,6 +344,162 @@
...
+}
+
+/* Two-tier ruler: major bands (month/year) on top, minor tick labels below.
+   Marker badges (today, milestones) are pinned here so they stay visible
+   while the marker line runs down the canvas. */
+.timelineAxis {
+  position: sticky;
+  top: 0;
+  z-index: 3;
+  height: 52px;
```
rohilsurana: **[minor / maintainability]** The `52px` axis height is a magic literal duplicated here and in `.timelineCanvas` (`min-height: calc(100% - 52px)`), plus the explaining comment. Change the axis height and you must remember all three. A custom property scoped to `.timelineRoot` (e.g. `--timeline-axis-height: 52px`) referenced in both rules would keep them in sync. Nice that it is not also duplicated in the JS.

packages/raystack/components/data-view/data-view.module.css
```diff
@@ -344,6 +344,162 @@
...
+.timelineGridline {
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  width: 0;
+  border-left: 0.5px dashed var(--rs-color-border-base-primary);
+  pointer-events: none;
+}
+
+.timelineMarkerLine {
```
rohilsurana, line 542: **[design question, low]** Marker lines, gridlines, and the cursor line render before the cards in DOM order and none carry a `z-index`, so cards paint on top of them. A `today` line reading as a subtle background is fine, but a custom `danger` marker (a deadline) would be hidden wherever a card overlaps its x. Is occluding marker lines behind cards intended? If milestone lines should stay visible, they would need to sit above cards (a higher `z-index`, still `pointer-events: none`).

rohilsurana (commented):
### Step 4: styles (`data-view.module.css`)

Checked the Timeline styles against house conventions. Tokens all resolve to real definitions in `styles/colors.css` and are the same ones the existing data-view CSS uses, so light/dark theming comes for free. The `0.5px` borders match the codebase convention (used in ~31 component stylesheets). Layering, sticky axis/footer, and `overscroll-behavior: contain` are all sensible. Two low-priority notes inline.

rohilsurana (commented):
### Step 5: tests (52 cases, `__tests__/timeline.test.tsx`)

This is a genuinely strong suite, well above what most feature PRs ship. It does not just assert "renders" — it checks exact geometry (px positions, widths, lane indices) and it exercises the hard, stateful machinery I was most worried about:

- momentum glide after a fling, plus wheel-cancels-glide, plus hold-still-then-release does *not* glide
- drag-to-pan on both axes, and correctly *not* panning from a card or a touch pointer
- binary-search horizontal culling under `virtualized` (overscan window)
- scroll anchoring: the left-edge time stays put when the domain is extended leftward
- `onVisibleRangeChange` dedup on identity churn, and re-fire on a real scroll
- scroll-position restore when a gated view is switched away and back
- the memoized card wrapper: a hover-cursor root re-render does not re-render card interiors
- the full `actionsRef` surface, including no-op-with-warning while hidden

Two gaps worth closing, both inline:

1. **`quarter` scale is untested** — and it is the only scale with hand-rolled (non-dayjs) logic, so it is the highest-value missing case.
2. **rAF is flushed with real timers**, which makes the momentum/timing tests the most likely to flake on CI and slows the suite.

One more, tied to my step-2 note: the `unitWidth <= 0` path (the potential hang) has no test because the guard does not exist yet. If the guard is added, a one-line test should come with it.

Net: coverage is not the concern here. The two items above are quality/robustness polish.

packages/raystack/components/data-view/__tests__/timeline.test.tsx
```diff
@@ -0,0 +1,976 @@
...
+    expect(ts.x(ts.t0)).toBe(0);
+    expect(ts.x(dayjs('2025-01-05').valueOf())).toBe(80);
+    const time = dayjs('2025-01-11').valueOf();
+    expect(ts.timeAt(ts.x(time))).toBe(time);
+    // Domain = Jan 1 → Feb 1 = 31 days.
+    expect(ts.totalWidth).toBe(31 * 20);
+  });
+});
+
+describe('buildAxis', () => {
```
rohilsurana, line 216: **[test coverage]** Every scale-level and renderer-level test uses `day` or `month`. The `quarter` scale has zero coverage, and it is the riskiest path because its logic is hand-rolled: `startOfUnit` (`date.startOf('month').subtract(date.month() % 3, ...)`), `addUnits` (`add(3 * n, 'month')`), and the `Q${...}` branches in `tickLabel` and `cursorLabel` all exist precisely because dayjs has no quarter support without a plugin. A single quarter-scale `buildAxis` case (assert `Q1 2025`/`Q2` ticks and year bands) would cover the one part of the math that is not just delegating to dayjs. A `week` case would be nice too, though it mostly delegates.

packages/raystack/components/data-view/__tests__/timeline.test.tsx
```diff
@@ -0,0 +1,976 @@
...
+    expect(root.dataset.dragging).toBeUndefined();
+    // Released — further movement no longer pans.
+    fireEvent.pointerMove(root, { pointerId: 1, clientX: 100, clientY: 0 });
+    await act(async () => {
+      await new Promise(resolve => setTimeout(resolve, 50));
+    });
+    expect(root.scrollLeft).toBe(50);
+  });
+
+  it('glides with momentum after a fling release', async () => {
```
rohilsurana, line 736: **[test robustness]** This suite flushes rAF by waiting real wall-clock time (`await new Promise(r => setTimeout(r, 30..100))`). It works, but the momentum/glide assertions here depend on rAF actually firing several times and the decay math producing a specific direction within that window, which is the classic recipe for a CI flake on a loaded machine, and it also makes the suite slower (many 30-100ms real waits add up across 52 tests). Consider `vi.useFakeTimers()` with a deterministic rAF stub (e.g. map rAF to a fake-timer tick) so `advanceTimersByTime` drives the glide exactly. Not blocking; the geometry tests are fine as-is, it is mainly the timing-dependent ones (this test and the cursor/scroll rAF ones) that would benefit.

apps/www/src/app/examples/timeline/page.tsx
```diff
@@ -0,0 +1,788 @@
...
+        runs.push([key, key]);
+      }
+    }
+    inflightRef.current += runs.length;
+    setIsLoading(true);
+    for (const [firstKey, lastKey] of runs) {
+      fetchOrdersApi(
+        dayjs(`${firstKey}-01`).valueOf(),
+        dayjs(`${lastKey}-01`).endOf('month').valueOf()
+      ).then(rows => {
```
rohilsurana: **[example quality] Confirming CodeRabbit's finding here, with a calibrated severity.** The observation is correct: this `.then()` has no `.catch()`/`.finally()`, and because the months are added to `requestedRef` at line 581 *before* the fetch settles, a rejection would strand `isLoading` at `true` and never retry that window.

But I would not call it Critical: `fetchOrdersApi` (line 161) only ever `resolve`s via `setTimeout`, so it cannot reject in this demo. The real reason to fix it is that the guide presents this block as *"the recommended pattern"* for range-window loading, and example code gets copied against a real API that can reject. So: worth fixing as a teaching example (add `.catch` that deletes the affected keys from `requestedRef` so a later scroll retries, plus `.finally` for the inflight decrement), but it is not a bug in the shipped library and does not block the feature.

apps/www/src/content/docs/components/dataview/props.ts
```diff
@@ -172,6 +172,175 @@ export interface DataViewListColumn {
...
+  /** Accessor key for the end date. Omitted → point markers; present → variable-width span cards. */
+  endField?: string;
+
+  /**
+   * Renders the card interior. The Timeline owns positioning (x from start,
+   * width from span, lane from packing, scroll); the consumer owns the card
+   * visual entirely. Compose `DataView.DisplayAccess` inside for Display
+   * Properties support. (Required)
+   */
+  renderCard: (row: T, context: TimelineCardContext) => ReactNode;
```
rohilsurana, line 206: **[docs] Re: CodeRabbit's "unbound `T`" finding here, I checked it and it is a false positive.** `T` is deliberately defined as `type T = any` at line 5 with a comment explaining it is a documentation placeholder so `auto-type-table` can render the API without real generics. So there is no compile error, and making the interface generic would fight the file's convention (see `DataViewListColumn.cell: (ctx: T)` right above, same pattern).

The bot's *other* point has a small kernel of truth: `renderCard`'s first arg is a TanStack `Row` wrapper, and consumers must read `row.original`. The generated table here shows `row: T`, which does not signal that. It is minor because (a) the file blurs all wrapper types to `T` by convention, and (b) the prose guide already shows `row.original` in every example. Optional polish: a one-line note in this description that the arg is a `Row` wrapper (data on `.original`).

rohilsurana (commented):
### Step 6: docs and demo (final pass)

The guide (`index.mdx`) is genuinely excellent: accurate to the real API (scale set, `unitWidth` defaults, `minCardWidth`, `estimatedPointWidth`, `actionsRef`, the range-window loading walkthrough), well-structured, and its code examples correctly use `row.original`. The `TimelineCardContext`/`TimelineMarker`/`TimelineActions` tables are wired up via `auto-type-table`, and the demo exports (`timelinePreview`, `timelinePointPreview`) resolve.

I also verified the two existing CodeRabbit findings rather than take them at face value:

- **`props.ts` "unbound T" (Major):** false positive. `T` is a deliberate `= any` placeholder (line 5). The only real sliver is that `renderCard`'s arg is a `Row` wrapper; minor, and the guide already teaches `row.original`. Details inline.
- **`page.tsx` unhandled rejection (Critical):** the code observation is right, but the mock never rejects, so it cannot happen here. Worth fixing only because the guide sells this as the recommended pattern and examples get copied. Downgraded to example-quality. Details inline.

Also worth noting for the author: the `estimatedRowHeight` doc here (line 263) already has to add "cards render at exactly this height" to undo the misleading name, which reinforces the naming point from my earlier types comment.

---

### Overall (steps 1-6 complete)

This is strong, careful work: a well-designed API consistent with `DataView.List` and `tour-beta`, solid scroll/pan/momentum and scroll-anchoring, correct and optimal lane packing, a comprehensive 52-test suite, on-pattern theming, and excellent docs. The data scrub this PR exists for is complete and clean.

Nothing I found blocks the concept. In priority order, before merge I would want:
1. the `unitWidth <= 0` guard in `time-scale.tsx` (only issue that can hang the render);
2. a `quarter`-scale test (the one untested hand-rolled path).

Everything else (the `estimatedRowHeight` name, `renderCard`/memo stability docs, keyboard a11y on the scroll region, the CSS 52px de-dup, marker-line z-order, real-timer test flake, and the example's fetch error handling) is polish the author can take or leave.

packages/raystack/components/data-view/components/timeline.tsx
```diff
@@ -0,0 +1,1061 @@
...
+  // unchanged window; each fire typically triggers a fetch check.
+  const lastNotifiedRangeRef = useRef<{ from: number; to: number } | null>(
+    null
+  );
+  useEffect(() => {
+    if (!onVisibleRangeChange || !viewport) return;
+    const from = timeScale.timeAt(viewport.left);
+    const to = timeScale.timeAt(viewport.left + viewport.width);
+    const prev = lastNotifiedRangeRef.current;
+    if (prev && prev.from === from && prev.to === to) return;
```
Shreyag02: The comment above says this dedupes "by ms window," but it's comparing raw decimals. When the domain shifts, the browser rounds the scroll position, so these never quite line up and the callback fires again — an extra fetch each time. Rounding both sides to whole ms before comparing would fix it.

apps/www/src/content/docs/components/dataview/props.ts
```diff
@@ -172,6 +172,175 @@ export interface DataViewListColumn {
...
+   * Hover crosshair snapped to the unit under the pointer, with a date badge on the axis.
+   * @defaultValue true
+   */
+  showCursorLine?: boolean;
+
+  /**
+   * Initial horizontal scroll target.
+   * @defaultValue "today"
+   */
+  defaultScrollTo?: TimelineDateInput | 'today' | 'start' | 'end';
```
Shreyag02, line 261: the rendered props table shows just TimelineDateInput for defaultScrollTo, so readers can't tell that 'today' | 'start' | 'end' are accepted — they only show up in the prose. 

Cause: `auto-type-table` resolves the type through the compiler, and since TimelineDateInput includes string, the literals get folded into it. Same for scrollTo's target (L331).

Runtime type is fine — this is purely cosmetic. 
Simplest fix is to name the values in the JSDoc, which survives extraction:

```
/**
 * Initial horizontal scroll target: a date, or `'today'` | `'start'` | `'end'`.
 * @defaultValue "today"
 */
defaultScrollTo?: TimelineDateInput | 'today' | 'start' | 'end';
```

Shreyag02:
Ordering and grouping doesn't work in the timeline mode so it should be disabled in that view

Shreyag02 (changes requested):

rohilsurana (approved):

Shreyag02 (approved):

PR #864, opened 2026-07-23
feat: add Reasoning AI element

rohanchkrabrty:
## Summary
- Adds the `Reasoning` AI-element component to `packages/raystack` — a collapsible thinking block built on `Collapsible` for displaying model reasoning/chain-of-thought in chat UIs.
- Supports a `streaming` prop that shows a shimmering "Thinking…" label and auto-opens the panel while reasoning streams in, auto-collapsing when done unless the user has toggled it manually, plus a `duration` prop that renders a "Worked for N seconds" label afterward.
- Ships CSS-module styling with `--rs-*` tokens and unit tests covering open/close, streaming auto-open/collapse, and controlled usage.
- Adds a Reasoning docs page (demos and props tables) under the AI Elements section of the website and exports the component from the package root.
- Stacked on #863 (`feat-ai-message`) — review that PR first; this branch adds one commit on top.

### DIFF
--- a/packages/raystack/components/reasoning/reasoning.tsx
+++ b/packages/raystack/components/reasoning/reasoning.tsx
@@ -1,0 +1,199 @@
+'use client';
+
+import { useControlled } from '@base-ui/utils/useControlled';
+import { ChevronRightIcon } from '@radix-ui/react-icons';
+import { cx } from 'class-variance-authority';
+import {
+  ComponentProps,
+  createContext,
+  ReactNode,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef
+} from 'react';
+import { Collapsible } from '../collapsible';
+import styles from './reasoning.module.css';
+
+interface ReasoningContextValue {
+  streaming: boolean;
+  duration?: number;
+  open: boolean;
+}
+
+const ReasoningContext = createContext<ReasoningContextValue | null>(null);
+
+function useReasoningContext(part: string): ReasoningContextValue {
+  const context = useContext(ReasoningContext);
+  if (!context) {
+    throw new Error(`Reasoning.${part} must be used within <Reasoning>`);
+  }
+  return context;
+}
+
+export interface ReasoningProps
+  extends Omit<ComponentProps<typeof Collapsible>, 'onOpenChange'> {
+  /**
+   * Whether the reasoning is still being produced. While `true` the default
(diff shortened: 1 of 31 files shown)


packages/raystack/components/reasoning/reasoning.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+ReasoningRoot.displayName = 'Reasoning';
+
+export interface ReasoningTriggerProps
+  extends ComponentProps<typeof Collapsible.Trigger> {}
+
+function defaultTriggerLabel(streaming: boolean, duration?: number): ReactNode {
+  if (streaming) {
+    return (
+      <span className={styles['reasoning-label-streaming']}>Thinking…</span>
```
paanSinghCoder, line 121: There is currently no way to overide this hardcoded text in isolation. User can only pass children. I think we can take this as prop and default can be `Thinking...`

packages/raystack/components/reasoning/reasoning.tsx
```diff
@@ -0,0 +1,199 @@
...
+    );
+  }
+  if (duration != null) {
+    return `Worked for ${duration} ${duration === 1 ? 'second' : 'seconds'}`;
```
paanSinghCoder, line 125: Same for this. IMO we can leave this for now though as this a composed string.

packages/raystack/components/reasoning/reasoning.tsx
rohanchkrabrty: User can pass `children` to override this string. The logic for `streaming` needs to come from outside so user's can handle the label accordingly

[REF](https://github.com/raystack/apsara/pull/864/changes#diff-01066c60f5053c3ee1b3c836b5c80ced3bd356cd753bff0be30a991ce6d19f94R141)

packages/raystack/components/reasoning/reasoning.tsx
rohanchkrabrty: User can pass `children` to override this string. The logic for `streaming` needs to come from outside so user's can handle the label accordingly

[REF](https://github.com/raystack/apsara/pull/864/changes#diff-01066c60f5053c3ee1b3c836b5c80ced3bd356cd753bff0be30a991ce6d19f94R141)

paanSinghCoder (approved):

PR #867, opened 2026-07-27
feat: add Sidebar.Trigger, collapsible modes (icon/hidden/none), and peekOnHover; fix a11y roles and label truncation

### LINKED ISSUE #840: Sidebar: header keeps opaque background in `inset` variant, and custom header content isn't hidden on collapse
### Summary

Two issues with the `Sidebar` component (`@raystack/apsara@1.0.1`).

---

### Bug 1 — `Sidebar.Header` keeps an opaque background in the `inset` variant

In the `inset` variant the sidebar root becomes transparent, but `Sidebar.Header` retains an opaque background. `Sidebar.Main` and `Sidebar.Footer` are transparent as expected — only the header paints a solid band. Because the header stretches the full width of the sidebar, this opaque band can paint over the edge/shadow of adjacent content (e.g. an inset content panel sitting flush against the sidebar).

**Steps to reproduce**
1. Render a `Sidebar` with `variant="inset"` containing `Sidebar.Header`, `Sidebar.Main`, and `Sidebar.Footer`.
2. Place a sibling element next to the sidebar that has a `box-shadow` along the edge adjacent to the sidebar.
3. Observe the top region where the header sits.

**Expected:** In `inset`, the header is visually transparent like the root / main / footer and does not cover adjacent content.

**Actual:** The header renders an opaque background and visibly covers the adjacent element's shadow along the header's height.

**Observation:** Among the sidebar sub-regions, only the header declares its own background. The `inset` variant resets the root's background but not the header's.

---

### Bug 2 — Custom `Sidebar.Header` content is not hidden when the sidebar is collapsed

When the sidebar is collapsed, built-in nav item labels and group labels are hidden correctly, but arbitrary content placed inside `Sidebar.Header` (e.g. an avatar + title) is not hidden. It is only clipped by the header's overflow, leaving a partial sliver of text visible.

**Steps to reproduce**
1. Render a collapsible `Sidebar` (`defaultOpen`) with a `Sidebar.Header` containing a leading icon/avatar plus a text title.
2. Collapse the sidebar.
3. Compare the header with the collapsed nav items.

**Expected:** Header content collapses cleanly, consistent with nav items (which fully hide their labels when collapsed).

**Actual:** A partial sliver of the header title text remains visible after collapse, while nav item labels disappear fully.

---

### Environment
- `@raystack/apsara`: `1.0.1`

ravisuhag:
## Summary

Started as an a11y/truncation audit of the Sidebar component and grew into a broader pass: it adds a `Sidebar.Trigger` subcomponent, two new collapse behaviors (hidden mode and `peekOnHover`), and a breaking prop cleanup that merges everything into a small API where every combination is meaningful.

### Breaking: `collapsible` is now a mode, not a boolean

- `collapsible?: 'icon' | 'hidden' | 'none'` (default `'icon'`) replaces the old boolean `collapsible` and the interim `collapseMode` prop.
  - `'icon'`: unchanged — shrinks to an icon rail, pushes content.
  - `'hidden'`: collapses to a thin edge strip and reveals as a floating panel with a backdrop when opened. Closes on backdrop click or Escape.
  - `'none'`: hides the resize handle and prevents toggling (was `collapsible={false}`).
- Deliberately a single prop rather than separate "can it collapse" and "collapsed shape" props — every value produces a sensible, unambiguous result instead of a matrix with invalid combinations.
- Known limitation (documented): the floating panel uses `position: fixed`, so it won't line up correctly under an ancestor with its own `transform`/`filter`/`contain`.
- Related renames: `hideCollapsedItemTooltip` → `hideItemTooltips` (it also hides the clipped-label tooltip when expanded, so the old name was wrong); the deprecated `tooltipMessage` prop is removed (use `collapseTooltip`); `useSidebar().collapsible` now mirrors the prop value instead of a boolean.

### New: `Sidebar.Trigger`

- A button that toggles the sidebar open/closed, works whether the Sidebar is controlled or uncontrolled — drop it into `Sidebar.Header` (or elsewhere) without wiring up `open`/`onOpenChange` yourself.
- Disabled automatically when `collapsible="none"`; a caller-passed `disabled` merges with that instead of overriding it.
- `useSidebar` and `SidebarContextValue` are now exported from the package for custom header/footer/item content that needs to read collapse state.
- A Trigger nested inside `Sidebar.Header`/`Main`/`Footer` gets hidden along with everything else when a `collapsible="hidden"` sidebar is closed — docs call this out and show the working pattern (Trigger as a direct child of `<Sidebar>`, or a separate toggle outside it entirely).

### New: `peekOnHover` and `peekDelay`

- Hovering a collapsed sidebar temporarily reveals it as the same floating panel, without touching the real `open` state — similar to VS Code's auto-hide sidebar. Reverts on mouse leave; clicking the resize handle or a `Sidebar.Trigger` pins it open for real.
- `peekDelay` (default 100ms) controls how long the pointer must hover before the peek starts.
- Consumer `onMouseEnter`/`onMouseLeave` handlers are forwarded, not overridden.
- No effect when `collapsible="none"` or while already open (documented).
- No keyboard equivalent in this pass — mitigated by existing per-item tooltips on focus, which already surface full labels to keyboard/screen reader users.

### Bug fix found during review

- During a peek, the sidebar's width expanded but item/group labels stayed invisible — several CSS rules hid content keyed on `data-closed` alone, which stays true throughout a peek (it tracks `open`, not `isPeeking`). Added floating-state overrides for every affected rule (`.nav-text`, `.nav-group-header`, `.nav-group-label`, `.nav-group-chevron`, `[data-collapse-hidden]`), placed after the hide rules so they win the cascade tie-break. Verified live in a browser via computed styles before/after the fix.

### Accessibility (original scope)

- `Sidebar.Main` is now `role="list"` (was `role="group"`); items render `role="listitem"`, which needs a list parent.
- Groups render as `role="listitem"` (was an unlabeled `<section>`, which mapped to a `region` landmark). Together with the inner `role="list"` items container, groups now form valid nested lists (`ul > li > ul`).
- Removed `role="banner"` from `Sidebar.Header` — banner is a page-level landmark and must not nest inside navigation.
- `aria-expanded` moved off the `aside` landmark onto the actual toggle controls (the resize handle and `Sidebar.Trigger`), which also set `aria-controls` pointing at the sidebar's id.
- Added: Escape closes the `collapsible="hidden"` floating panel; documented that the floating panel doesn't trap focus or set `aria-modal`.

### Text truncation (original scope)

- Group labels and more-menu labels now truncate with an ellipsis, matching item labels, with `min-width: 0` so ellipsis reliably engages inside flex rows.
- Expanded items whose label is clipped now show the full text in a tooltip on hover/focus (previously only collapsed items got this).

### Cleanup

- Consolidated identical typography declarations across `.nav-text`, `.nav-group-label`, `.more-menu-item-text` into one shared rule.
- `white-space: nowrap` → `text-wrap: nowrap`, matching button/breadcrumb/filter-chip.
- Docs site nav updated to the new `collapsible="none"` API.

## Testing

- 66/66 sidebar unit tests pass (new coverage for `Sidebar.Trigger`, collapse modes, `peekOnHover`, `peekDelay`); 2162/2162 across the full package.
- Verified in the docs site in a real browser: roles are correct in the DOM, truncation and tooltips behave as described, the `collapsible="hidden"` floating panel opens/closes via backdrop click and Escape, and `peekOnHover` expands with fully visible labels and reverts on mouse-leave (confirmed via computed styles, not just visual inspection).


packages/raystack/components/sidebar/sidebar-root.tsx
```diff
@@ -6,33 +6,94 @@ import {
...
+  collapseTooltip?: ReactNode;
+  /** @deprecated Renamed to `collapseTooltip`; will be removed in the next
+   *  major version. Ignored when `collapseTooltip` is set. */
   tooltipMessage?: ReactNode;
   open?: boolean;
   defaultOpen?: boolean;
   onOpenChange?: (open: boolean) => void;
 }
 
+const PEEK_OPEN_DELAY = 100;
```
rohanchkrabrty: This should be a prop `peekDelay` in SidebarRoot

packages/raystack/components/sidebar/sidebar-root.tsx
```diff
@@ -57,52 +125,106 @@ export function SidebarRoot({
...
+          className={cx(styles.root, className)}
+          data-position={position}
+          data-variant={variant}
+          data-open={visualOpen ? '' : undefined}
+          data-closed={!visualOpen ? '' : undefined}
+          data-collapse-disabled={!collapsible ? '' : undefined}
+          data-collapse-mode={collapseMode}
+          data-peeking={isPeeking ? '' : undefined}
+          onMouseEnter={handleMouseEnter}
+          onMouseLeave={handleMouseLeave}
```
rohanchkrabrty, line 219: If consumers pass `onMouseEnter` and `onMouseLeave` , our handlers will be overriden.

Let's destructure them in the props, and call the provided functions inside the `handleMouseEnter/Leave`.
Eg -`onMouseEnter?.()`

rohanchkrabrty (changes requested):

rohanchkrabrty (approved):

PR #871, opened 2026-07-31
feat: improve Amount formatting capabilities

### LINKED ISSUE #595: [Amount] Improve formatting capabilities and fix edge cases
## Summary

Improvements to the Amount component to enhance formatting capabilities, fix edge cases, and follow component conventions.

### 2. Add `notation` prop
No way to get compact formatting like `$1.2M` or `$13K`. `Intl.NumberFormat` supports `notation: 'compact'` natively. This is useful for dashboards and summary views.

### 3. Add `narrowSymbol` to `currencyDisplay`
`Intl.NumberFormat` supports `'narrowSymbol'` (e.g., `$` instead of `US$` in non-US locales), but the `currencyDisplay` prop only allows `'symbol' | 'code' | 'name'`.

### 4. Add `signDisplay` prop
`Intl.NumberFormat` supports `signDisplay` (`'auto'`, `'always'`, `'exceptZero'`, `'never'`) for controlling `+`/`-` sign display — useful for showing gains/losses.

### 5. Memoize formatter instances
`getCurrencyDecimals()` and `isValidCurrency()` each create a new `Intl.NumberFormat` instance on every render. These should be memoized or combined into a single formatter creation.

### 6. Fix string slicing edge case for short strings
`value.slice(0, -decimals) + '.' + value.slice(-decimals)` breaks for strings shorter than the decimal count (e.g., `value="5"` with USD → `".5"` instead of `"0.05"`).

## Moved to Cross-Cutting Issues

The following items have been moved to cross-cutting issues that address them across all components:

- Forward HTML attributes (#674)
- TypeScript strictness (#675)

ravisuhag:
## Summary

Implements the remaining items from #595, plus a `tabularNums` prop.

### New props
- **`notation`** (`'standard' | 'compact'`, default `'standard'`) — compact formatting like `$1.2M` / `$13K` for dashboards and summary views. Note: compact rounds aggressively by design (Intl behavior), so small values like `$12.99` render as `$13`.
- **`currencyDisplay: 'narrowSymbol'`** — always shows the narrow symbol (`$` instead of `US$` in non-US locales).
- **`signDisplay`** (`'auto' | 'always' | 'exceptZero' | 'never'`, default `'auto'`) — controls the `+`/`-` sign, useful for gains/losses.
- **`tabularNums`** (default `true`) — fixed-width figures so digits align across table rows. Previously this was always on via CSS; it is now a prop so running text can opt out for proportional figures. Default behavior is unchanged.

### Performance
- `Intl.NumberFormat` instances are now memoized in a module-level cache (keyed by locale + options, capped at 64 entries). Previously every render created 3 formatter instances, which adds up fast when Amount renders in table rows.
- `isValidCurrency()` and `getCurrencyDecimals()` are merged into a single cached `getCurrencyInfo()` lookup. Decimals now come from `resolvedOptions().maximumFractionDigits` instead of regex-matching formatted output.

### Not included
- Issue item 6 (string slicing edge case) is already fixed on main — the `padStart` conversion handles short strings correctly (`"5"` → `$0.05`).

### Docs
- Playground controls for the new props, plus `notation`, `signDisplay`, and `tabularNums` example sections.

## Test plan
- 15 new test cases covering all four props and their interactions (compact + hideDecimals / string values / hideCurrency / explicit fraction digits; sign preserved when hideCurrency strips the currency token; tabular class toggling).
- Full Amount suite: 64/64 passing. Lint and typecheck clean for the touched files.

Closes #595

### DIFF
--- a/apps/www/src/content/docs/components/amount/demo.ts
+++ b/apps/www/src/content/docs/components/amount/demo.ts
@@ -32,8 +32,18 @@ export const playground = {
     },
     currencyDisplay: {
       type: 'select',
-      options: ['symbol', 'code', 'name'],
+      options: ['symbol', 'narrowSymbol', 'code', 'name'],
       defaultValue: 'symbol'
+    },
+    notation: {
+      type: 'select',
+      options: ['standard', 'compact'],
+      defaultValue: 'standard'
+    },
+    signDisplay: {
+      type: 'select',
+      options: ['auto', 'always', 'exceptZero', 'never'],
+      defaultValue: 'auto'
     },
     minimumFractionDigits: {
       type: 'number',
@@ -50,6 +60,10 @@ export const playground = {
     hideCurrency: {
       type: 'checkbox',
       defaultValue: false
+    },
+    tabularNums: {
+      type: 'checkbox',
+      defaultValue: true
     }
   },
   getCode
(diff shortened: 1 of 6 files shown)


packages/raystack/components/amount/amount.tsx
```diff
@@ -269,14 +335,14 @@ export const Amount = ({
         );
 
     return (
-      <span {...props} className={cx(styles.amount, className)}>
+      <span {...props} className={cx(tabularNums && styles.tabular, className)}>
```
paanSinghCoder, line 338: `tabularNums` only toggles `font-variant-numeric`, so this is CSS-only and doesn't need a prop. We deliberately removed styling props like Input's `width`.

apps/www/src/content/docs/components/amount/demo.ts
paanSinghCoder, line 133: `narrowSymbol` example is missing here.

packages/raystack/components/amount/amount.tsx
```diff
@@ -52,9 +52,27 @@ export interface AmountProps extends ComponentProps<'span'> {
   /**
    * Currency display format
    * @default 'symbol'
-   * @example 'symbol' - $12.99, 'code' - USD 12.99, 'name' - 12.99 US Dollars
+   * @example 'symbol' - $12.99 (may show "US$" in non-US locales), 'narrowSymbol' - $12.99 (always the narrow symbol), 'code' - USD 12.99, 'name' - 12.99 US Dollars
```
paanSinghCoder, line 55: Nitpick - It should be 12.99 US _dollars_ and not _Dollars_.

Issue #848, opened 2026-07-11
DataView: forward align on Filters/DisplayControls, and let Search control its leading icon


ravisuhag:
## Summary

Two `DataView` controls hard-code a positioning / rendering detail and expose no prop to change it, even though the underlying primitive already supports it.

1. `DataView.Filters` and `DataView.DisplayControls` — popup alignment is fixed.
2. `DataView.Search` — the leading magnifier icon can't be removed or replaced.

## 1. Forward `align` on Filters and DisplayControls

- `DataView.DisplayControls` exposes `trigger` + `hideViewSwitcher`/`hideOrdering`/`hideGrouping`/`hideDisplayProperties`, but no `align`.
- `DataView.Filters` exposes `trigger` + `classNames`, but no `align`.

The underlying `Menu.Content` / `Popover.Content` already extend Base UI's `Positioner.Props` and accept `align` / `side` / `sideOffset`; these controls just don't forward them.

**Request:** expose `align` (and ideally `side` / `sideOffset`) on both `DataView.Filters` and `DataView.DisplayControls`, forwarding to the underlying menu / popover. Default to today's behavior. Base UI's Positioner already handles collision flip/shift, so this is prop-forwarding only.

## 2. Let `Search` control its leading icon

`DataView.Search` extends the base `Search`, whose props are `Omit<InputProps, 'leadingIcon'>`. The base `Input` accepts `leadingIcon`, but `Search` removes it and hard-codes a magnifier, with no supported way to hide or replace it.

**Request:** re-expose `leadingIcon` on `Search` (settable to a custom node or `null`), or add a `showSearchIcon={false}` / `hideIcon` flag. Default keeps the magnifier.

## Willing to contribute

Happy to open a PR if the approach looks right.

Issue #846, opened 2026-07-11
DataView: expose classNames for filter-chip internals and loader rows


ravisuhag:
## Summary

`DataView.List` renders its rows, cells, and group headers internally, and exposes `classNames` hooks for them (`root`, `header`, `headerCell`, `row`, `cell`, `groupHeader`, plus per-column `cell`/`header` and `Filters.classNames.filterChips`/`addFilter`). Two rendered parts still have no hook, so theming them means targeting hashed CSS-module classes (`[class*="data-view-module_…"]`), which are internal and can change between releases.

## What has no hook

1. **Filter-chip internals.** `Filters.classNames.filterChips` lands on the chip root only. The inner parts can't be reached: `chip-label`, `operationText`, `operationValue`, `selectValue`, `removeIcon`, `removeIconContainer`, `leadingIcon`.

2. **Loader / skeleton rows.** When `isLoading` renders placeholder rows, there's no hook on the loader container or its rows — `DataViewListClassNames` covers real rows, not the loading state.

## Proposed change

1. Add a `classNames` object to `FilterChip` (surfaced through `Filters`) covering its inner parts.
2. Add a `loader` (and/or loader-row) key to `DataViewListClassNames`.

Both follow the pattern the existing hooks already use: read an optional string from props, apply it to the element `DataView` already renders.

## Willing to contribute

Happy to open a PR if the approach looks right.

ravisuhag:
@rohanchkrabrty Can you check if this issue looks good?

rohanchkrabrty:
All issues are resolved. Closing this

PR #872, opened 2026-07-31
feat: add data-slot attributes across the entire design system

ravisuhag:
## Summary

Every element every component renders now carries a stable `data-slot` attribute — a kebab-case, component-prefixed identifier (`filter-chip-remove`, `data-view-list-loader-row`, `dialog-close`). This gives consumers a supported hook for styling and testing inner parts without targeting hashed CSS-module classes.

Started as a 3-component pilot (Search, FilterChip, DataView) and grew into full coverage across the design system per follow-up discussion.

Related: #846 (closes its styling half — filter-chip internals and loader rows are now reachable). The prop-forwarding half of #848 is separate and unaffected.

## Coverage

**76 of 77 components** get slots. `data-table` is deliberately excluded — it's slated for deprecation in favor of `DataView`, so it doesn't get new public API surface.

Every component ships with:
- `data-slot` on every element it renders, placed before any `{...props}` spread so a caller-supplied override always wins
- a `__tests__/data-slots.test.tsx` asserting the slot contract via the shared `expectSlots`/`getSlot`/`getAllSlots` helper (`packages/raystack/test-utils/data-slots.ts`), including that conditional slots are absent when their part isn't rendered
- a `### Slots` section on its docs page listing slot names

`DataView.List` cells also carry `data-column={accessorKey}`, so a single column can be targeted without a per-column `classNames` entry: `[data-slot="data-view-list-cell"][data-column="amount"]`.

## Convention

Documented in the styling guide (*With data-slot*, including the `has-[]`/`[&_]` composition patterns) and enforced going forward by the `add-new-component` skill checklist:
- kebab-case, component-prefixed: `filter-chip-remove`, `data-view-list-row`
- subparts extend the prefix; a repeated part reuses one name
- slot names are public API, covered by semver
- `Popover.Content`/`Menu.Content`/`Dialog.Popup`-style wrappers that split props between a Base UI Positioner and Popup get the slot hardcoded directly on the Popup — not routed through forwarded props, since those land on the positioner instead
- existing internal precedent (`menu-trigger`'s `data-slot='menu-subtrigger'`) is preserved unchanged
- a component reaching into a *different* component's rendered markup by tag name (`.container input`, `.root button`) is exactly the fragile pattern `data-slot` replaces — the 7 places this existed in-scope are now `[data-slot=...]` selectors instead

## classNames cleanup (closing the loop on #846)

#846 asked to grow `classNames` with more keys (a filter-chip object, a loader key). Instead we audited every `classNames`-style prop in the design system against the now-complete `data-slot` coverage:

- **Deprecated** (`@deprecated` JSDoc, non-breaking — actual removal is a future major version) every key with an exact 1:1 `data-slot` match: `DataView.List`, `DataView.Timeline`, `DataView.Filters` (`filterChips`/`addFilter`), `Indicator`, `Input`, `EmptyState`, `Table.SectionHeader`, `Sidebar.NavigationGroup`, `Sidebar.Item`, and `Sidebar.More`'s `root`/`leadingIcon`/`text`.
- **Left alone, deliberately:** `Sidebar.More.classNames.menuContent` — `Menu.Content` portals to `document.body`, so a `data-slot` selector can't be scoped to one specific dropdown instance; this prop remains the only way to target it. Same reasoning applies to any other portaled per-instance content.
- **Fixed a real bug:** `DataView.Filters.classNames.addFilter` was declared in the type but never applied anywhere — setting it silently did nothing. Wired it onto the default trigger.
- **Closed two doc gaps** found while cross-referencing: `Table` had no `### Slots` section at all, and `DataView.List`/`Timeline`/`Sidebar.NavigationGroup`'s `classNames` types were undocumented or oversimplified in the docs-site mirror (`apps/www/.../props.ts`).

## Fixes made while integrating

A few real issues surfaced while wiring this up across the whole system, fixed as part of this PR:
- **calendar**: a test wrongly assumed `DatePicker`'s error span unmounts when there's no error — it always mounts and toggles via `data-visible`, so screen readers hear it the instant an error appears. Fixed the test, not the component.
- **avatar**: a test asserted the fallback and loaded-image slots simultaneously — Base UI unmounts the fallback once the image loads. Split into two tests.
- **chat**: a test rendered `Chat.JumpButton` outside its required `Chat.Messages` ancestor. Fixed the test harness.
- **text, headline, label, grid, grid-item**: Base UI's `mergeProps` types its parameter against `HTMLAttributes`, which doesn't include `data-slot` — passing it as a fresh object literal fails TypeScript's excess-property check. Fixed by extracting the props into a named `const` first (matching the existing pattern already used in `flex.tsx`), which sidesteps the check without a cast.

## Testing

- Full package suite: **2389 tests passing** (1 pre-existing skip), across all 76 covered components
- Typecheck: unchanged from the pre-existing baseline (6 errors, all in `data-table`/`menu`/`popover` test files, none introduced by this change)
- `@raystack/apsara` rollup build: passes
- `www` docs site build: passes
- Biome: clean (only formatting auto-fixes applied)


packages/raystack/components/tour/tour-overlay.tsx
rohanchkrabrty, line 205: This div is missing `data-slot='tour-overlay-hit'`

rohanchkrabrty (approved):
Following component docs are missing `Slots` section -

accordion, announcement-bar, avatar, calendar, callout, code-block, collapsible, color-picker, combobox, empty-state, field, fieldset, form, list, scroll-area, and select.

PR #874, opened 2026-08-01
feat(ci): add pkg.pr.new canary releases

ravisuhag:
## Summary
- Add `.github/workflows/canary.yaml` to publish a preview build of `@raystack/apsara` to [pkg.pr.new](https://pkg.pr.new) on every PR and push to `main`, so changes can be installed and tested before a real npm release
- Document the canary workflow in `CONTRIBUTING.md`
- Add a "Try a preview build" section to the Getting Started docs

## Test plan
- [ ] Merge and confirm the workflow runs on the next PR
- [ ] Confirm pkg.pr.new comments on the PR with an install link
- [ ] Install the preview build in a test project and verify it works

### DIFF
--- a/.github/workflows/canary.yaml
+++ b/.github/workflows/canary.yaml
@@ -1,0 +1,41 @@
+name: Canary Release
+
+# Publishes a preview build of @raystack/apsara to pkg.pr.new on every push
+# to a pull request or to main, so changes can be installed and tested
+# before a real npm release. See https://pkg.pr.new
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+permissions: {}
+
+jobs:
+  publish:
+    name: Publish to pkg.pr.new
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v4
+
+      - name: Setup pnpm 9
+        uses: pnpm/action-setup@v4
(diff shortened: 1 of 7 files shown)


.github/workflows/canary.yaml
```diff
@@ -0,0 +1,49 @@
...
+          cache: "pnpm"
+
+      - name: Install Dependencies 🔧
+        run: pnpm install --frozen-lockfile
+
+      - name: Build Step 🔧
+        run: pnpm build:apsara
+
+      - name: Publish 🚀
+        run: pnpm exec pkg-pr-new publish ./packages/raystack --comment=update
```
rohanchkrabrty: Let's use the `--packageManager=pnpm` flag to shown `pnpm` in the generated comments

rohanchkrabrty: Also I think we should use `--previewVersion` 


> Use --previewVersion to rewrite every published package's version to 0.0.0-preview-<sha> before packing (working tree is restored afterward):
>
>pnpm exec pkg-pr-new publish --previewVersion './packages/*'
For example, 1.0.0 becomes 0.0.0-preview-a832a55. The 0.0.0- prefix cannot satisfy a normal semver range for a real release, so lockfile collisions with npm versions are avoided.

From their [docs^](https://github.com/stackblitz-labs/pkg.pr.new#setup)

rohanchkrabrty (approved):

Issue #578, opened 2026-02-02
[Theme] Enhance customization capabilities


ravisuhag:
## Summary

Enhance the ThemeProvider to support comprehensive theming customization, making Apsara more adaptable to different design requirements and use cases.

## Current Capabilities

The ThemeProvider already handles:
- Theme mode (`light` / `dark` / `system`)
- Style variants (`modern` / `traditional`)
- Accent colors (`indigo` / `orange` / `mint`)
- Gray scales (`gray` / `mauve` / `slate` / `sage`)
- Persistence via localStorage
- System preference detection
- FOUC prevention with inline script

### 1. Scaling/Density
```typescript
scaling: 'compact' | 'default' | 'comfortable' | 'spacious'
```
Adjusts spacing tokens globally for different UI densities. Useful for data-dense applications vs content-focused interfaces.

### 2. Border Radius Control
```typescript
radius: 'none' | 'small' | 'medium' | 'large' | 'full'
```
Global control over component roundness beyond style variants. Allows sharp corners for enterprise apps or pill-shaped components for playful designs.

### 3. Typography Scale
```typescript
fontFamily: { body: string; heading: string; mono: string }
fontSize: 'small' | 'default' | 'large'  // Base scale multiplier
```
Allow consumers to customize font families and adjust the base font size scale.

### 4. Custom Token Overrides
```typescript
tokens: {
  colors?: Partial<ColorTokens>
  spacing?: Partial<SpacingTokens>
  radius?: Partial<RadiusTokens>
  shadows?: Partial<ShadowTokens>
}
```
Allow consumers to override any design token for brand-specific customization.

### 5. Component-Level Defaults
```typescript
components: {
  Button?: { variant?: 'solid' | 'outline'; size?: 'sm' | 'md' }
  Input?: { variant?: 'default' | 'soft' }
  // etc.
}
```
Set default props for components at the theme level, reducing repetition across the app.

### 6. Animation Control
```typescript
reducedMotion: boolean | 'system'
transitionDuration: 'none' | 'fast' | 'default' | 'slow'
```
Respect user accessibility preferences and allow global animation speed control.

### 7. Nested Appearance Mode
```typescript
appearance: 'light' | 'dark' | 'inherit'  // Per-subtree theming
```
Allows nested theme contexts with different modes (e.g., dark sidebar in light app).

### 8. Panel Background Mode
```typescript
panelBackground: 'solid' | 'translucent'  // For glassmorphism effects
```
[...]

PR #880, opened 2026-08-02
feat: RFC theme improvements

rohanchkrabrty:
### Summary

- Proposes rewriting `Theme` to mount design tokens on a theme element instead of writing attributes to `<html>`, so the root theme, a scoped subtree, and a portalled popover all become the same mechanism — collapsing today's `Root`/`Scoped` split, making the root server-renderable, and fixing scoped theming for the 13 portalling components where it is currently broken.
- Adds the customisation surface tracked in #578: `appearance`, `accentColor`, `grayColor` (with `auto` pairing), `radius`, `scaling`, `panelBackground`, `reducedMotion`, and font families — behind a `defaultValue`/`value` controlled-per-key API with opt-out persistence.
- Adds per-component `radius` overrides with element-only, non-compounding semantics (a component prop affects only that component; tree-level changes stay with `Theme`), via a shared cva variant and CSS module.
- Restructures the token stylesheets: `:where()` on every `--rs-*` declaration so consumer overrides win regardless of load order, a stable `.rs-theme` override target, spacing/effects/z-index moved off `:root` so they become scopable, a factor-based radius scale, and a new surface token family behind `panelBackground`.
- Retires the `style` prop, audits all 10 items in #578 with verdicts, and records the rejected options — component-level defaults, cursor tokens, a `tokens` deep-override prop, and keeping tokens on `<html>` — with rationale. Breaking, with no shim; a migration guide accompanies implementation.

### DIFF
--- a/docs/rfcs/004-theme-improvements.md
+++ b/docs/rfcs/004-theme-improvements.md
@@ -2,7 +2,7 @@
 ID: RFC 004
 Created: August 3, 2026
 Status: Draft
-RFC PR: TBD
+RFC PR: https://github.com/raystack/apsara/pull/880
 ---
 
 # Theme Improvements
--- a/docs/rfcs/004-theme-improvements.md
+++ b/docs/rfcs/004-theme-improvements.md
@@ -9,7 +9,7 @@ RFC PR: https://github.com/raystack/apsara/pull/880
 
 This RFC proposes rewriting the `Theme` component and the token stylesheets under it. Tokens move off `<html>` onto a real element, so the root theme, a nested scope, and a portal all work the same way. That fixes server rendering, the one provider per page limit, and scoped themes breaking inside portals.
 
-On that foundation it adds the customisation surface tracked in [issue #578](https://github.com/raystack/apsara/issues/578), appearance, accent, gray, radius, scaling, panel background, reduced motion and font families, together with per-component `radius` overrides, and it retires `style`. This is a breaking change: no compatibility shim ships, and a migration guide accompanies the implementation.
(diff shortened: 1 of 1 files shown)


ravisuhag:
One thought on `fontFamily` — since it's not persisted and there's no picker for it (per the "What Persists" section), maybe it doesn't need to be a component prop at all. It's a one-time branding choice, so plain CSS variables would do the job just as well, and it'd sidestep an issue: it's applied as an inline style, which always beats the `:where()`-wrapped token rules, so it's the one setting that can't be overridden by consumer CSS the way everything else can.

If we later want a real font switcher (persisted, changeable at runtime), that's worth designing properly then, with font loading and metrics correction in scope. For now, keeping it as plain CSS vars seems like less to build and maintain.

ravisuhag:
Question on multiple root themes: `color-scheme` on `<html>` is document-wide, it can only be one value at a time. If two independent root Themes exist on a page (say a host app plus an embedded widget, both marked `data-rs-root`) with different `appearance` values, which one wins? Worth spelling out whether only the outermost root gets to touch `<html>`, or if that setup just isn't supported.

ravisuhag:
With multiple root Themes now supported, I think there's a collision risk in persistence: storage is per-setting-key under a prefix (`storageKey`), and most consumers won't think to set a custom one unless they know they need to. Two independent root Themes both using the default prefix would read/write the same localStorage entries and overwrite each other's saved settings. Might be worth either documenting this clearly as a requirement, or having each root auto-generate a unique key if none is passed.

ravisuhag (approved):

rohanchkrabrty:
**1. `fontFamily`**

Agreed, dropped the prop. Fonts are now just `--rs-font-body/title/mono`, overridable from `.rs-theme` like every other token.

**2. Multiple roots / `color-scheme`**

Right now it was depending on stylesheet ordering. I have added `isRoot={false}` so a theme provider can opt out, and alsodocumented the fallback. Nested scopes were never affected, they still get their own `color-scheme` from the existing `[data-theme]`

**3. Storage collision**

The shared-key case was feature where themes could stay in sync, which is what you'd want for several live examples on a docs page.

Auto-generating key would create more problems, so persistence is now opt-in. Persistence is enabled if we pass `persistKey`.

@ravisuhag

PR #881, opened 2026-08-04
fix: datepicker to emit onErrorChange instead of built-in error UI

rohanchkrabrty:
## Summary

- Add `onErrorChange?: (error: string | undefined) => void` to `DatePicker` — fires on typed-input validity transitions only (a message when the text stops parsing as a valid in-bounds date, `undefined` when it recovers or the picker commits/closes), mirroring the `onOpenChange` convention.
- Remove DatePicker's built-in error presentation — the inline "Invalid date" `role="alert"` span, its reserved-line CSS, and the `aria-invalid`/`aria-describedby` wiring. Error display is now consumer-owned: lift the callback into `Field`'s `error` prop (which also wires `aria-invalid` via Base UI) or a form library. Breaking for consumers relying on the built-in message.
- Track the error in a ref instead of state (nothing renders from it anymore), which also removes the stale-error closure read in the outside-click commit path.
- Rework tests to assert through `onErrorChange` and add coverage for transition-only firing, clearing on recovery/commit, and a no-error-UI regression test.
- Docs: document `onErrorChange` in the props reference and add a "With Field" live example to the DatePicker demos.


apps/www/src/content/docs/components/calendar/demo.ts
```diff
@@ -184,6 +184,26 @@ export const datePickerDemo = {
...
+      description="Type a date or pick one from the calendar"
+      error={error}
+      style={{ maxWidth: 240 }}
+    >
+      <DatePicker
+        onErrorChange={setError}
+        slotProps={{ input: { size: "medium" } }}
+      />
+    </Field>
+  );
```
rohilsurana, line 204: do we need to call the function here to render it? or the framework takes care of that?

packages/raystack/components/calendar/calendar.module.css
```diff
@@ -343,27 +343,3 @@
 .datePickerInput {
   text-align: left;
 }
-
-.datePickerError {
```
rohilsurana, line 347: `filter-chip.module.css` has a rule that hides the DatePicker error line inside a chip:
```
.dateFieldWrapper [class*="datePickerError"] { display: none; }
```

we can probably get rid of that too.

rohilsurana (approved):

PR #883, opened 2026-08-07
feat: icon provider with lucide replacing radix icons

rohanchkrabrty:
## Summary

- Apsara draws its icons with `lucide-react` instead of `@radix-ui/react-icons`. It is a peer dependency (`>=0.500.0 <1.0.0`), so the app picks the version. Some icons inside Apsara's components therefore draw a different shape — the migration guide lists them.
- The package exports 31 icons: the ones Apsara's own components draw. Each has a stable key that does not name the library it comes from, such as `SearchIcon`, `SortAscendingIcon` or `ClearIcon`. All of them render at 16×16 with `strokeWidth={1.5}` and set `data-icon="<Key>"`, so CSS can target a single icon.
- **Breaking:** `@raystack/apsara/icons` used to export raw in-house SVG components. Twelve of those names are gone — `BellIcon`, `BellSlashIcon`, `BuildingsFilledIcon`, `CheckCircleFilledIcon`, `CoinIcon`, `CoinColoredIcon`, `CrossCircleFilledIcon`, `OrganizationIcon`, `ResetIcon`, `ShoppingBagFilledIcon`, `SidebarIcon`, `TriangleRightIcon` — and the glyph comes from `lucide-react` instead. `CoPilotIcon` is unchanged; `FilterIcon` keeps its name with a new drawing.
- `createIcon` is exported, so an app can build its own icons the same way and they behave the same way — same props, same `data-icon`, same overriding.
- To change an icon, pass `<Theme icons={{ components, props }}>`: `components` swaps a drawing by key, `props` applies to every icon built with `createIcon`. Maps are partial, and nested `<Theme>`s layer key by key. `Select.Trigger`'s own `iconProps` prop is not affected.
- `icons/icons.tsx` is a normal source file — one `createIcon` call per icon — and `IconName` is derived from its exports, so the type cannot drift from the code. `icons/__tests__/bundle.test.ts` bundles a small fixture with the package's own rollup config and asserts the icons it does not use are gone from the output, which is what lets one file hold all of them. There is no icon codegen, no icon build step and no CI check for either.
- `apps/www` imports `lucide-react` directly for any glyph the package does not export, and the Icons docs are now one page. The changelog entry and the lucide migration guide describe the shipped API.


ravisuhag:
Out of 243 icons only 26 of icons are used by Apsara's own components. The other 217 are a curated catalog for consumers, and the curation is what creates a problem: lucide has 1773 icons, so anyone needing one outside our 243 has to either PR `icon-map.json` and wait for a release, or import from lucide directly and hand-set `size={16} strokeWidth={1.5}`. Mixed usage is guaranteed in any real app, and every name we export is API we have to keep stable. The "swap libraries with a one-file edit" promise also gets harder the bigger the catalog is, since a future library won't have equivalents for everything.

**Proposal: ship 26, export the factory**

1. Ship only the 26 icons the components use (25 lucide + CoPilot).

2. Make `createIcon` public from `@raystack/apsara/icons`, with the name loosened to accept any string. Then consumers build their own map in their app, and that file gives them everything we were trying to give them with the catalog:

```tsx
// src/icons.ts — the app's single place for icons
import { createIcon } from '@raystack/apsara/icons';
import { Rocket, Trash2 } from 'lucide-react';

export const RocketIcon = createIcon('Rocket', Rocket);
export const TrashIcon = createIcon('Trash', Trash2);
```

Correct sizing by default, `data-icon` for CSS, one-file swap later — same promise we keep for ourselves, now in their hands. And there's no cliff: the 1773-icon question disappears because the catalog is theirs.

3. Split the Theme API by job:
   - `Theme icons` stays, but typed to the 26 built-in names only. Its real purpose is reskinning Apsara's own components — users can't reach those call sites any other way, and this is what makes white-labeling and the radix-look migration map work. (Narrow typing is safe: we can widen later without breaking, but can't narrow back.)
   - `Theme iconProps` applies to all icons, including user-created ones, since they read the same context. That's the one thing that must stay global or "tune stroke width in one place" breaks.

**Implementation: at 26 icons, we can drop the codegen too**

The map + codegen design earns its keep at 243 icons. At 26 it's machinery we don't need: the JSON map, the generator script, gitignored generated files, the `prepare`/`prebuild`/`predev`/`pretest` hooks, and the `check:icon-map` CI script can all go. One committed file replaces them:

```tsx
// icons/icons.tsx — this file IS the map
import { Check, ChevronDown, X /* ...23 more */ } from 'lucide-react';
import { createIcon } from './create-icon';

export const CheckIcon = createIcon('CheckIcon', Check);
export const ChevronDownIcon = createIcon('ChevronDownIcon', ChevronDown);
// ...
```

Every property of the current design survives:

- **Library swap stays a one-file edit** — this file is the map now, and editing 26 lines by hand is a five-minute job. Codegen's value was doing that 243 times.
- **Tree-shaking is unchanged** for ESM via `/*#__PURE__*/`. The only regression is CJS consumers loading 26 tiny wrappers instead of one — negligible at this size, unlike at 243.
- **The CI check becomes unnecessary.** With real imports in a committed file, a dropped lucide export fails `tsc` and fails the existing render tests. Ordinary tooling catches what the custom script exists to catch.
- CoPilot becomes a plain hand-written component wrapped with `createIcon` — no SVG asset pipeline.

And we gain: icon code is visible in the repo and reviewable in PRs, `clone && test` works with no generation step, and there's no "forgot to regenerate" failure mode. If the catalog ever grows past ~50, the map + codegen design is the right tool to bring back — it's a good design, just for a bigger problem than 26 icons.

**Where this goes later: native multi-library support falls out for free**

If we ever want Apsara to officially support two or three icon libraries (say lucide and radix) and let people switch through Theme, this architecture already is that feature — a "native" library is just a pre-made, tested override map promoted from user-space into the package. The PR's own migration guide proves it: its copy-paste radix map is exactly this file, just living in the docs.

Each library ships as one committed file behind its own entry point, exporting a set — icons and their tuning together, since libraries don't share prop semantics (lucide is stroke-based in a 24-unit box, radix is fill-based in 15):

```tsx
// @raystack/apsara/icons/radix
export const radixIconSet: IconSet = {
  icons: { CheckIcon: RadixCheck, XIcon: RadixX /* ...24 more */ },
  iconProps: { width: 16, height: 16 } // no strokeWidth: silent on what radix doesn't need,
                                       // so lucide-based custom icons keep their stroke default
};
```

Theme grows one prop, and switching becomes:

```tsx
import { radixIconSet } from '@raystack/apsara/icons/radix';

<Theme iconSet={radixIconSet}>
  <App />
</Theme>
```

Each alternate library is an optional peer dependency — apps that never import the set never pull it into the bundle. And because a set is plain data, the existing layering keeps working: per-name `icons` overrides win over the set, `iconProps` merges on top, and a nested `<Theme iconSet={...}>` can run a different library for one section of the app.

At 26 names, a third library is 26 hand-written lines and a render test. At 243 it would be another codegen target and another 243 names to reconcile on every swap — one more reason the small catalog is the version of this design that scales in the direction we'd actually grow.

**Docs: one page in the Theme section**

The current docs are shaped around the big catalog — an Icons section with a Usage page and a searchable 243-icon gallery. With 26 icons and one recipe, the content collapses to a single "Icons" page, and it belongs in the Theme section, since after this change everything interesting about icons flows through Theme:

1. **Basics** — components use 26 named icons, lucide-backed by default; import and use them directly.
2. **Your own icons** — the `createIcon` + `src/icons.ts` recipe. The main section, since it's what most readers come for. Link to lucide.dev to browse — 1773 icons with better search than we'd ever build.
3. **Switching sets** — `<Theme iconSet={radixIconSet}>`, the optional peer dep, per-name overrides winning over the set, `iconProps` merging, nested Themes. Plus a "roll your own set" one-liner.
4. **The 26 built-in names** — an inline grid, each icon labeled with its override name, toggleable between the two shipped sets. Replaces the gallery page; its job is "look up the name of the icon you want to replace", not "find an icon to use".
5. **API reference** — `createIcon`, `IconProvider`, `iconSet`/`icons`/`iconProps`, types.

The migration guide stays its own document (release-specific, will age out), and gets shorter: "restore the radix appearance" becomes one line with `iconSet`, plus one new step — "if you imported an icon we no longer ship, add it to your own `icons.ts` with `createIcon`". The 600-line gallery component and the Icons nav section go away.

**What we give up:** the ready-made catalog and the big gallery page. The docs become the 26 built-in names plus the recipe above. In exchange the public API drops from 243 names to 26 plus one factory, users own their icon set with no gap, and the next library migration touches a fraction of the surface.

ravisuhag:
We should also use lucide sparkle icon instead of custom Copilot.

rohanchkrabrty:
@ravisuhag Let's keep the catalog and the codegen.

Going from 243 to 29 doesn't save much. Codegen writes the icons, `check:icon-map` validates them, and icons only ship when imported. The gallery builds itself from the barrel, so none of this scales with icon count.

#### Catalog

The catalog is also an opinionated choice on our side. We're intentionally defining a stable, curated set of public icon names rather than exposing Lucide's API directly. The key is keeping those 243 public names stable, which the registry solves. If Lucide drops or renames something, CI catches it and it's a one-line JSON change.

#### createIcon

Agreed on making createIcon public. If we don't ship an icon today, you lose the default size, data-icon, and override support. A public factory with a plain string name fixes that and gives us both the catalog and the flexibility to create custom icons.

#### iconSet
I don't think we need it. `<Theme icons iconProps>` already covers this, and adding a third prop just creates more interaction to reason about.

ravisuhag (approved):

Shreyag02 (approved):

PR #886, opened 2026-08-10
feat(kbd): add Kbd component, remove Command.Shortcut

Shreyag02:
## Description

Adds `Kbd`, a component for displaying keyboard keys and shortcut sequences.

Apsara has no shared way to render a keyboard key today, so it gets hand-rolled with inline styles wherever it appears. `Command.Shortcut` rendered unstyled `<kbd>` elements, `Menu` has no shortcut support, and tooltips have no way to show an action's shortcut.

```tsx
<Kbd>Esc</Kbd>

<Kbd.Group>
  <Kbd>⌘</Kbd>
  <Kbd>K</Kbd>
</Kbd.Group>
```

The API follows [shadcn's Kbd](https://ui.shadcn.com/docs/components/base/kbd), converted to Apsara conventions: dot-notation sub-components and `--rs-*` tokens. The visual treatment reuses the existing key styling from `Command.Shortcut` and the docs-site `Kbd`, so no new tokens were added.

Two variants: `solid` (default) and `ghost`, for surfaces that already carry their own background — a command item, a tooltip, an input. Setting `variant` on `Kbd.Group` applies it to the keys inside; a key's own `variant` still wins. That inheritance is resolved in CSS with `:where()`, the same way `radio.module.css` handles group-level size, so no React context is involved.

One note for review: `Kbd.Group` renders a `<kbd>` rather than a `<div>`, since the HTML spec defines a nested `kbd` as an individual key within a larger input.

### Breaking: `Command.Shortcut` is removed

Per review, `Command.Shortcut` is removed rather than re-pointed at `Kbd`. It existed to split a whitespace-separated string into one `<kbd>` per token, and it was only ever used as an item's `trailingIcon` — where a `Kbd` or `Kbd.Group` says the same thing without the loop.

```diff
-<Command.Item trailingIcon={<Command.Shortcut>⌘ K</Command.Shortcut>}>
+<Command.Item
+  trailingIcon={
+    <Kbd.Group variant="ghost">
+      <Kbd aria-label="Command">⌘</Kbd>
+      <Kbd>K</Kbd>
+    </Kbd.Group>
+  }
+>
```

The `command-shortcut` and `command-shortcut-key` data-slots go with it; style the keys via `kbd` / `kbd-group` instead. `Command.Shortcut` ships in `@raystack/apsara@1.5.0`, so this needs to land in the release notes as a breaking change — the commit carries a `BREAKING CHANGE:` footer with the migration above.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Improvement (Improvements to existing code)
- [ ] Other (please specify)

## How Has This Been Tested?

- 25 tests for `Kbd` covering rendering, `className` merging, ref forwarding, variants, group inheritance, composition, and the `data-slot` contract. Full suite: 2597 passed, 1 skipped.
- `pnpm build:apsara` passes and the Vercel preview builds the docs site.
- `tsc --noEmit` and `biome check` clean on all changed files.
- Variant cascade verified against the *built* stylesheet (hashed class names, real theme tokens): a bare key is solid, a `ghost` group ghosts its keys, an explicit `variant="solid"` inside a ghost group stays solid, and keys ignore pointer events throughout.
- Checked in light and dark themes on a tooltip surface, in a command item, and inline in text.

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (.mdx files)
- [x] I have added tests that prove my fix is effective or that my feature works

N/A

## Related Issues

N/A

### DIFF
--- a/apps/www/src/content/docs/components/kbd/demo.ts
+++ b/apps/www/src/content/docs/components/kbd/demo.ts
@@ -1,0 +1,87 @@
+'use client';
+
+export const preview = {
+  type: 'code',
+  code: `<Kbd.Group>
+    <Kbd>⌘</Kbd>
+    <Kbd>K</Kbd>
+  </Kbd.Group>`
+};
+
+export const singleDemo = {
+  type: 'code',
+  code: `<Flex gap={5} align="center">
+    <Kbd>Esc</Kbd>
+    <Kbd>⌘</Kbd>
+    <Kbd>⇧</Kbd>
+    <Kbd>↵</Kbd>
+    <Kbd>Tab</Kbd>
+  </Flex>`
+};
+
+export const groupDemo = {
+  type: 'code',
+  code: `<Flex gap={7} align="center">
+    <Kbd.Group>
+      <Kbd>⌘</Kbd>
+      <Kbd>K</Kbd>
+    </Kbd.Group>
+    <Kbd.Group>
+      <Kbd>⌘</Kbd>
+      <Kbd>⇧</Kbd>
+      <Kbd>P</Kbd>
+    </Kbd.Group>
+  </Flex>`
+};
+
+export const separatorDemo = {
+  type: 'code',
+  tabs: [
+    {
+      name: 'Plus',
+      code: `<Kbd.Group>
+        <Kbd>⌘</Kbd>
+        +
+        <Kbd>K</Kbd>
+      </Kbd.Group>`
+    },
+    {
+      name: 'Then',
+      code: `<Kbd.Group>
+        <Kbd>G</Kbd>
+        then
+        <Kbd>P</Kbd>
+      </Kbd.Group>`
+    }
+  ]
+};
+
+export const withTextDemo = {
+  type: 'code',
+  code: `<Flex gap={3} align="center">
+    <Text size="small" variant="secondary">Press</Text>
+    <Kbd.Group>
+      <Kbd>⌘</Kbd>
+      <Kbd>K</Kbd>
+    </Kbd.Group>
+    <Text size="small" variant="secondary">to open the command palette</Text>
+  </Flex>`
+};
+
+export const withTooltipDemo = {
+  type: 'code',
+  code: `<Tooltip>
+    <Tooltip.Trigger render={<Button variant="outline" />}>
+      Search
+    </Tooltip.Trigger>
+    <Tooltip.Content>
+      <Flex gap={3} align="center">
+        Open search
+        <Kbd.Group>
+          <Kbd>⌘</Kbd>
+          <Kbd>K</Kbd>
+        </Kbd.Group>
+      </Flex>
+    </Tooltip.Content>
+  </Tooltip>`
+};
--- a/apps/www/src/content/docs/components/kbd/demo.ts
+++ b/apps/www/src/content/docs/components/kbd/demo.ts
@@ -1,21 +1,52 @@
 'use client';
 
-export const preview = {
-  type: 'code',
-  code: `<Kbd.Group>
-    <Kbd>⌘</Kbd>
-    <Kbd>K</Kbd>
-  </Kbd.Group>`
+import type { ComponentPropsType } from '@/components/demo/types';
+import { getPropsString } from '@/lib/utils';
+
+export const getCode = (props: ComponentPropsType) => {
+  const { children, ...rest } = props;
+
+  return `<Kbd${getPropsString(rest)}>${children}</Kbd>`;
+};
+
+export const playground = {
+  type: 'playground',
+  controls: {
+    variant: {
+      type: 'select',
+      options: ['solid', 'ghost'],
+      defaultValue: 'solid'
+    },
+    children: {
+      type: 'text',
+      initialValue: 'Esc'
+    }
+  },
+  getCode
 };
 
 export const singleDemo = {
   type: 'code',
   code: `<Flex gap={5} align="center">
     <Kbd>Esc</Kbd>
-    <Kbd>⌘</Kbd>
-    <Kbd>⇧</Kbd>
-    <Kbd>↵</Kbd>
+    <Kbd aria-label="Command">⌘</Kbd>
+    <Kbd aria-label="Shift">⇧</Kbd>
+    <Kbd aria-label="Enter">↵</Kbd>
     <Kbd>Tab</Kbd>
+  </Flex>`
+};
+
+export const variantDemo = {
+  type: 'code',
+  code: `<Flex gap={7} align="center">
+    <Kbd.Group>
+      <Kbd aria-label="Command">⌘</Kbd>
+      <Kbd>K</Kbd>
+    </Kbd.Group>
+    <Kbd.Group variant="ghost">
+      <Kbd aria-label="Command">⌘</Kbd>
+      <Kbd>K</Kbd>
+    </Kbd.Group>
   </Flex>`
 };
 
@@ -23,12 +54,12 @@ export const groupDemo = {
   type: 'code',
   code: `<Flex gap={7} align="center">
     <Kbd.Group>
-      <Kbd>⌘</Kbd>
+      <Kbd aria-label="Command">⌘</Kbd>
       <Kbd>K</Kbd>
     </Kbd.Group>
     <Kbd.Group>
-      <Kbd>⌘</Kbd>
-      <Kbd>⇧</Kbd>
+      <Kbd aria-label="Command">⌘</Kbd>
+      <Kbd aria-label="Shift">⇧</Kbd>
       <Kbd>P</Kbd>
     </Kbd.Group>
   </Flex>`
@@ -40,7 +71,7 @@ export const separatorDemo = {
     {
       name: 'Plus',
       code: `<Kbd.Group>
-        <Kbd>⌘</Kbd>
+        <Kbd aria-label="Command">⌘</Kbd>
         +
         <Kbd>K</Kbd>
       </Kbd.Group>`
@@ -58,14 +89,17 @@ export const separatorDemo = {
 
 export const withTextDemo = {
   type: 'code',
-  code: `<Flex gap={3} align="center">
-    <Text size="small" variant="secondary">Press</Text>
-    <Kbd.Group>
-      <Kbd>⌘</Kbd>
-      <Kbd>K</Kbd>
-    </Kbd.Group>
-    <Text size="small" variant="secondary">to open the command palette</Text>
-  </Flex>`
+  code: `<Text size="small" variant="secondary">
+    Press <Kbd.Group><Kbd aria-label="Command">⌘</Kbd><Kbd>K</Kbd></Kbd.Group> to open the command palette.
+  </Text>`
+};
+
+export const withInputDemo = {
+  type: 'code',
+  code: `<Input
+    placeholder="Search projects"
+    trailingIcon={<Kbd variant="ghost" aria-label="Command K">⌘K</Kbd>}
+  />`
 };
 
 export const withTooltipDemo = {
@@ -77,8 +111,8 @@ export const withTooltipDemo = {
     <Tooltip.Content>
       <Flex gap={3} align="center">
         Open search
-        <Kbd.Group>
-          <Kbd>⌘</Kbd>
+        <Kbd.Group variant="ghost">
+          <Kbd aria-label="Command">⌘</Kbd>
           <Kbd>K</Kbd>
         </Kbd.Group>
       </Flex>
--- a/apps/www/src/content/docs/components/kbd/index.mdx
+++ b/apps/www/src/content/docs/components/kbd/index.mdx
@@ -74,7 +74,7 @@ Use `Kbd` on its own for a one-key hint. Keys share a minimum width so a narrow 
 
 ### Sequences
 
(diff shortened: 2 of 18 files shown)


packages/raystack/components/kbd/kbd.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+KbdRoot.displayName = 'Kbd';
+
+export type KbdGroupProps = ComponentProps<'kbd'>;
+
+/**
+ * Renders a `<kbd>` rather than a `<div>`: per the HTML spec a `kbd` nested
+ * inside a `kbd` represents an individual key within a larger input, which is
+ * exactly a shortcut sequence.
+ */
```
rohanchkrabrty: Please remove comments

packages/raystack/components/kbd/kbd.module.css
```diff
@@ -0,0 +1,32 @@
...
+
+.kbd,
+.kbd-group {
+  display: inline-flex;
+  align-items: center;
+  color: var(--rs-color-foreground-base-tertiary);
+  font-family: var(--rs-font-body);
+  font-size: var(--rs-font-size-mini);
+  line-height: var(--rs-line-height-mini);
+  letter-spacing: var(--rs-letter-spacing-mini);
```
rohanchkrabrty: font color and styles should be applied only on `.kbd` and not the group so that there is a difference of text between two sets of .kbd

packages/raystack/components/kbd/kbd.tsx
rohanchkrabrty, line 1: Create 2 variants
- solid (default)
```
background: var(--rs-color-background-neutral-primary);
color: color: var(--rs-color-foreground-base-secondary);
```
- ghost
```
background: transparent;
color: color: var(--rs-color-foreground-base-tertiary);
```

packages/raystack/components/kbd/kbd.module.css
```diff
@@ -0,0 +1,32 @@
...
+}
+
+.kbd {
```
rohanchkrabrty, line 10: Should have `pointer-events: none` and `user-select: none`

packages/raystack/components/kbd/kbd.module.css
```diff
@@ -0,0 +1,32 @@
...
+  justify-content: center;
+  box-sizing: border-box;
+  height: var(--rs-space-6);
+  /* Square minimum so a narrow "K" reads the same width as a wide "⌘". */
```
rohanchkrabrty: Remove comments

packages/raystack/components/kbd/kbd.module.css
```diff
@@ -0,0 +1,32 @@
+/* normalize.css sets `kbd { font-family: monospace }`, so both parts restore
+   the body font explicitly rather than relying on inheritance. */
```
rohanchkrabrty: Remove comments

apps/www/src/content/docs/components/kbd/demo.ts
rohanchkrabrty, line 1: Add examples of usage in Input

apps/www/src/content/docs/components/kbd/index.mdx
```diff
@@ -0,0 +1,96 @@
+---
+title: Kbd
+description: Displays a keyboard key or a shortcut sequence.
```
rohanchkrabrty: A component for displaying keyboard keys and shortcuts.

apps/www/src/content/docs/components/kbd/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+### Root
+
+A single keyboard key. Renders a `<kbd>` element.
+
+<auto-type-table path="./props.ts" name="KbdProps" />
+
+### Group
+
+Spaces a sequence of keys evenly. Also renders a `<kbd>`: per the HTML spec, a `kbd` nested inside a `kbd` represents an individual key within a larger input, which is exactly what a shortcut sequence is.
```
rohanchkrabrty: Groups multiple keyboard keys for key combinations.

packages/raystack/components/kbd/kbd.module.css
```diff
@@ -0,0 +1,32 @@
...
+  line-height: var(--rs-line-height-mini);
+  letter-spacing: var(--rs-letter-spacing-mini);
+}
+
+.kbd {
+  justify-content: center;
+  box-sizing: border-box;
+  height: var(--rs-space-6);
+  /* Square minimum so a narrow "K" reads the same width as a wide "⌘". */
+  min-width: var(--rs-space-6);
```
rohanchkrabrty, line 14: Also add `width: fit-content` or else it will stretch in certain cases

apps/www/src/content/docs/components/kbd/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+### In a tooltip
+
+A common use is surfacing a shortcut alongside the action it triggers.
+
+<Demo data={withTooltipDemo} />
+
+## Accessibility
+
+- `Kbd` is presentational and renders the semantic `<kbd>` element, which screen readers announce as keyboard input.
```
rohanchkrabrty: `which screen readers announce as keyboard input.` is misleading here, remove it

apps/www/src/content/docs/components/kbd/index.mdx
rohanchkrabrty, line 1: Let's add a playground too since we are introducing variants now

apps/www/src/content/docs/components/kbd/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+`Kbd.Group` renders whatever you put between the keys, so separators are plain text. Use `+` for keys pressed together and a word like `then` for keys pressed in order.
+
+<Demo data={separatorDemo} />
+
+### Inline with text
+
+Keys sit on the text baseline, so they can be dropped into a sentence.
+
+<Demo data={withTextDemo} />
```
rohanchkrabrty, line 91: Even tho the example says Inline with text, the actual example is rendering in side a Flex, so it's misleading

packages/raystack/components/kbd/kbd.tsx
rohanchkrabrty, line 1: There is a sub-component in `Command` named `Command.Shortcut` 
That should now be an alias of `Kbd` component with the default as `ghost`. All props should be forwarded

rohanchkrabrty (changes requested):

apps/www/src/content/docs/components/kbd/props.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export interface KbdProps {
+  /** The key to display, e.g. `⌘`, `Esc`, or `Enter`. */
+  children?: ReactNode;
+
+  /**
+   * Visual style variant. Inherited from a parent `Kbd.Group` when set there.
+   * @defaultValue "solid"
+   */
+  variant?: 'solid' | 'ghost';
```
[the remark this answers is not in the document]
Shreyag02: Going to leave the prop tables as they are here.

These `props.ts` files under `apps/www/src/content/docs` aren't the component's real types — they're hand-written mirrors that exist only to feed `auto-type-table`. The convention across the docs site is to list component-specific props and let `className` stand in as the escape hatch: of the 69 `props.ts` files under `content/docs/components`, none extends a native React type, none documents a `ref`, and the single file that references those types at all goes the other way deliberately:

```ts
export type AmountProps = Omit<ApsaraAmountProps, keyof ComponentProps<'span'>>;
```

Inheriting `ComponentProps<'kbd'>` would render a table of every HTML attribute for a component with three real props, which is presumably why `Amount` strips them back out.

The forwarding is documented in prose instead, in the API Reference intro:

> Both parts render a `<kbd>` element and forward any native attributes (`id`, `title`, `aria-label`, …) to it.

And the published types are the real contract — `KbdProps` and `KbdGroupProps` are both `ComponentProps<'kbd'> & VariantProps<typeof kbd>`, and `CommandShortcutProps` is `KbdGroupProps` — so consumers get full attribute and ref type-checking from the package regardless of what the docs table lists. Ref forwarding is covered by tests in `kbd.test.tsx` and `command.test.tsx`.

Happy to revisit if the docs site ever moves to generating tables from the package types directly, since that would want a consistent answer across all 69 files rather than just this one.

packages/raystack/components/kbd/kbd.tsx
```diff
@@ -1,28 +1,51 @@
...
-const KbdGroup = ({ className, ...props }: KbdGroupProps) => (
-  <kbd
-    data-slot='kbd-group'
-    className={cx(styles['kbd-group'], className)}
-    {...props}
-  />
+export type KbdGroupProps = ComponentProps<'kbd'> & VariantProps<typeof kbd>;
+
+const KbdGroup = ({ className, variant, ...props }: KbdGroupProps) => (
+  <KbdGroupContext.Provider value={variant ?? undefined}>
```
rohanchkrabrty: Why do we need context here? Let's not use context in a simple component like KBD.

The variant at group level can be solved using css only. If it can't be done let's remove the support and keep the variant at kbd level only.

packages/raystack/components/command/command-misc.tsx
```diff
@@ -63,11 +64,12 @@ export const CommandSeparator = ({
 };
 CommandSeparator.displayName = 'Command.Separator';
 
-export type CommandShortcutProps = ComponentProps<'span'>;
+export type CommandShortcutProps = KbdGroupProps;
```
rohanchkrabrty: Let's remove the `Command.Shortcut` component. I missed that it used to map over a string.

Keeping that complexity of loop an valid element check doesn't make sense especially when it's just used as a trailingIcon. 
```
<Command.Item trailingIcon={<Command.Shortcut>⌘ B</Command.Shortcut>}
```

rohanchkrabrty (changes requested):

packages/raystack/components/kbd/kbd.tsx
Shreyag02: i will check

Shreyag02: Context is gone. `Kbd.Group` now just puts a variant class on the group and the inheritance is resolved in CSS with `:where()` — the same trick `radio.module.css` uses for group-level size, where the group rule sits low enough in specificity that a key's own `variant` still wins. Solid moved onto the base `.kbd`, so a bare key needs no variant class at all.

Verified against the built stylesheet (hashed class names, real theme tokens) rather than just the source: a bare key is solid, a `ghost` group ghosts its keys, and `<Kbd variant="solid">` inside a ghost group stays solid.

packages/raystack/components/command/command-misc.tsx
Shreyag02: Removed. `Command.Item` takes a `Kbd` / `Kbd.Group` as `trailingIcon` now, so the string-splitting loop and the `isValidElement` branch both go away:

```tsx
<Command.Item
  trailingIcon={
    <Kbd.Group variant="ghost">
      <Kbd aria-label="Command">⌘</Kbd>
      <Kbd>K</Kbd>
    </Kbd.Group>
  }
>
```

The `command-shortcut` and `command-shortcut-key` data-slots go with it — style the keys via `kbd` / `kbd-group` instead. Docs, demos, the slot table, and the tests all follow.

Worth flagging for the release notes: `Command.Shortcut` ships in `@raystack/apsara@1.5.0`, so this is a public breaking change. The PR title carries a `!` and the commit has a `BREAKING CHANGE:` footer with the migration above, so it still reads as breaking after the squash merge.

Shreyag02:
@rohanchkrabrty rest of the earlier batch is all in — GitHub still shows the threads as unresolved, so here's the mapping if you want to bulk-resolve them:

| Your comment | Where it landed |
|---|---|
| Remove comments (`kbd.tsx`, `kbd.module.css` ×2) | Both files carry no comments now |
| `pointer-events: none` + `user-select: none` | `kbd.module.css` — verified as computed `pointer-events: none` on keys |
| `width: fit-content` | `kbd.module.css`, on both the key and the group |
| Two variants, solid default | `solid` = `background-neutral-primary` / `foreground-base-secondary`, `ghost` = `transparent` / `foreground-base-tertiary` — the ghost text colour was on `secondary` for a while, it's on `tertiary` as you specced now |
| Font/colour on `.kbd` only, not the group | Group is `gap` + `font: inherit`; typography and colour live on `.kbd`. Separator text between keys takes the surrounding size (16px vs the key's 11px) |
| Docs description | Frontmatter uses your wording verbatim |
| Group description | Same |
| Drop "which screen readers announce as keyboard input" | Gone |
| Add a playground | Added, with a `variant` select |
| "Inline with text" rendered inside a `Flex` | That example renders inside `Text` now |
| Add Input usage examples | Added — single `ghost` key, since the input's trailing slot is a fixed 16×16 box and a multi-key group gets clipped at the border |

Also worth knowing: the docs demos give symbol-only keys a name (`<Kbd aria-label="Command">⌘</Kbd>`), since the bare glyph carries no name for the key it stands for. I've reworded the Accessibility notes too — they were predicting how screen readers read `⌘`/`⇧`/`↵`, which is the same over-claim you had me strip from the `<kbd>` line earlier.

rohanchkrabrty (approved):

PR #889, opened 2026-08-18
feat(data-view): lane per sort value in the timeline

rsbh:
## Description

Adds a third timeline lane packing mode: **one lane per value of the sorted-by field**.

```tsx
// Sorting the "High"/"Medium"/"Low" label alphabetically gives High, Low, Medium,
// so carry a numeric rank and sort on that.
<DataView
  data={tasks}
  fields={fields}
  defaultSort={{ name: 'rank', order: 'asc' }}
  getRowId={t => t.id}>
  <DataView.Timeline
    startField="start"
    endField="end"
    lanePacking="one-per-sort-value"
    renderCard={renderCard}
  />
</DataView>
```

Rows sharing a value share a lane, packed by date within it; a value only claims a **sub-lane** where two of its own cards genuinely overlap in time. So a timeline sorted by priority rank reads as a High lane, a Medium lane and a Low lane, and only the priority with concurrent work grows a second row.

The active sort does double duty: it picks the field lanes are built from *and* orders them. That means one vocabulary rather than a parallel set of lane props, and the Ordering control rebuilds lanes live.

### Behaviour details

- **Lane field and order** both come from `tableQuery.sort[0]`. No sort in the query → falls back to `auto` (the root requires `defaultSort`, so this is a guard, not a mode).
- **Rows with no usable value** — null, `undefined`, `""`, or a non-primitive (which also logs a dev warning) — share one lane, always last, wherever the sort would have placed them.
- **Values are keyed by their string form**, so `1` and `"1"` share a lane.
- **With `group_by` active**, each section gets its own lane set and `context.laneIndex` stays section-relative. Grouping by the sorted field is allowed and degenerates cleanly: a section already holds one value, so it renders as one lane plus sub-lanes on overlap.

### Also: declared section order

`DataViewField.groupOrder?: string[]` ranks group sections for **every** renderer that groups — `['High', 'Medium', 'Low']`, which text sorting can't produce. Values it doesn't list follow in first-occurrence order.

⚠️ **Behaviour change, no opt-in required:** rows with no value now land in the **last** section instead of wherever their bucket was first seen. `groupData` previously emitted `groupMap.forEach`, i.e. insertion order. Anyone grouping a `DataView.List` (or a timeline) on a nullable field will see that section move, whether or not they declare `groupOrder`. `group-data.test.ts` pins both the declared and undeclared paths.

Sections and timeline lanes share only the empty-bucket-last half of the rule (`orderBucketKeys`): sections rank by `groupOrder`, lanes by the sort. One field that is both grouped and sorted can therefore order its sections and its lanes differently — that's the contract of `one-per-sort-value`, and both call sites say so.

### New API

| Surface | Addition |
| --- | --- |
| `DataViewField` | `groupOrder?: string[]` |
| `DataViewTimelineProps` | `lanePacking: 'auto' \| 'one-per-row' \| 'one-per-sort-value'` |

Additive apart from the null-section ordering change called out above.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update
- [x] Test (adding missing tests or correcting existing tests)

## How Has This Been Tested?

- 41 new tests: `order-bucket-keys.test.ts` (9) and `group-data.test.ts` (7) for section ordering, 11 in `pack-lanes.test.ts` for `packLanesBySortValue` (bucketing, sub-lane splits, differential against `packLanes` for a single bucket, plus randomized no-overlap-within-a-lane and lanes-never-span-buckets invariants), 14 in `timeline.test.tsx` (lane assignment from the sort, direction flip, relaning when the sort field changes, rank-field ordering, dotted `accessorKey` paths, a sort key matching no field, null and non-primitive values, per-section lanes under `group_by`, virtualized culling).
- Full package suite green (**228 data-view tests**, 2677 in the package); no pre-existing test modified.
- Lane values are read via `row.getValue(...)`, so a dotted `accessorKey` (`meta.rank`) lanes on the same value TanStack sorted by; a sort key with no matching field warns instead of silently collapsing.
- `tsc --noEmit` clean on every file this branch touches, in the package and in `apps/www` (the repo's other pre-existing errors are unchanged).
- Verified live on the docs site: the demo lanes High (2 lanes — one from a real overlap) → Medium (3) → Low (2) under `rank asc`, no group bands, console clean.

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (.mdx files)
- [x] I have added tests that prove my fix is effective or that my feature works

Docs → DataView → Timeline → **Lane packing** carries a live demo (one lane per priority, sorted by rank, Ordering control left visible) alongside a mode-comparison table.

## Related Issues

n/a


packages/raystack/components/data-view/components/timeline.tsx
```diff
@@ -535,7 +552,19 @@ export function DataViewTimeline<TData>({
...
           if (endTime !== null && endTime < startTime) endTime = startTime;
         }
-        items.push({ row, startTime, endTime });
+        // Lane bucket — the sorted-by field's value. Resolved here so packing
+        // below is pure geometry. Only a primitive identifies a lane; anything
+        // else (an object, an array) shares the no-value lane rather than
+        // collapsing into one "[object Object]" bucket.
+        let laneKey: string | null = null;
+        if (fieldLanes) {
+          const value = original?.[laneField as string];
```
Shreyag02: A dotted `accessorKey` silently flattens the whole timeline
 
```ts
const value = original?.[laneField as string];
if (value == null || value === '') laneKey = null;
```
 
This reads the field off the raw object. But TanStack understands dots as a path — `@tanstack/table-core/core/column.js:31` does `accessorKey.split('.')`. And `fieldsToColumnDefs` hands it your key unchanged (`utils/index.tsx:73`).
 
So with `accessorKey: 'meta.rank'`:
 
- **Sorting works.** TanStack digs into `meta.rank` and sorts correctly.
- **Laning doesn't.** `original['meta.rank']` is `undefined`, so every row is treated as "has no value" and they all pile onto one lane.
I ran it with three rows of ranks 1/2/3. The list sorted correctly (`B C A`), and the timeline gave me:
 
```
lanes → { a: '0', b: '0', c: '0' }
```
 
The really unpleasant part: **no warning fires.** The check on line 563 only catches objects and functions. `undefined` goes down the `value == null` path on 562, so `unlaned` never increments and the warning on 583 stays quiet. I spied on `console.warn` — zero calls. So the timeline just looks wrong with nothing telling you why.
 
**Fix:** `row` is the loop variable on line 543, so it's already available:
 
```ts
const value = row.getValue(laneField as string);
```
 
That gives you exactly what TanStack sorted on. It's also safe if the sort name has no matching column — `getValue` returns `undefined` rather than throwing (`table-core/core/row.js:27-36`).
 
I applied this locally and it works: ranks 1/2/3 land on lanes 0/1/2, and all 97 existing timeline tests still pass.
 
If you'd rather not change the lookup, then at least warn when *every* row comes back unlaned. Nobody means to do that.

packages/raystack/components/data-view/utils/pack-lanes.tsx
```diff
@@ -47,6 +48,61 @@ export function packLanes(
...
+    let bucket = buckets.get(key);
+    if (!bucket) {
+      bucket = [];
+      buckets.set(key, bucket);
+    }
+    bucket.push(index);
+  }
+
+  let laneCount = 0;
+  for (const key of orderBucketKeys([...buckets.keys()])) {
```
Shreyag02, line 93: Lanes ignore `groupOrder`, so the same field can order two different ways

```ts
for (const key of orderBucketKeys([...buckets.keys()])) {
```
 
No second argument, so no declared order ever reaches lane packing. Compare with `utils/index.tsx:123`, where `groupData` *does* pass it:
 
```ts
for (const key of orderBucketKeys([...groupMap.keys()], field?.groupOrder)) {
```
 
So if a field declares `groupOrder: ['High', 'Medium', 'Low']` and is also the sorted-by field, you get two different orders out of one field:
 
| | what you see |
| --- | --- |
| Lanes (`one-per-sort-value`) | High, Low, Medium ← alphabetical, from the sort |
| Sections (`group_by` on that field) | High, Medium, Low ← from `groupOrder` |
 
I checked both against what actually renders.
 
This is also the exact scenario the comment in **`order-bucket-keys.tsx:19-21`** says can't happen:
 
> Shared by `groupData` … and `packLanesBySortValue` … so sections and lanes can never disagree about where a value sits.
 
**Two ways to resolve it, either is fine:**
 
**a) Let lanes honour it too.** Pass the declared order down from the call site at `timeline.tsx:736`:
 
```ts
packLanesBySortValue(items, undefined, field?.groupOrder)
```
 
Then the comment becomes true, and the demo's `priority` field works on its own — no need for the parallel `rank` field.
 
**b) Say lanes follow the sort, full stop.** Then drop the `orderBucketKeys` call here entirely. With no `order` passed it's only doing "keep input order, put the empty bucket last", which is two lines inline. And fix the comment.
 
Right now it's neither — it *looks* like the two paths share an ordering rule, but they only share half of one. That's a leftover from `eb0b3d3`, back when `laneOrder` was still a prop.

packages/raystack/components/data-view/components/timeline.tsx
```diff
@@ -698,15 +732,26 @@ export function DataViewTimeline<TData>({
...
+                  width: item.packWidth
+                }))
+              );
       const entry = { ...section, ...packed, laneOffset: offset };
       offset += packed.laneCount;
       return entry;
     });
     return { laidOutSections: list, laneCount: offset };
-  }, [positionedSections, lanePacking]);
+  }, [positionedSections, lanePacking, fieldLanes]);
```
Shreyag02:
```ts
}, [positionedSections, lanePacking, fieldLanes]);
```
 
`laneField` isn't in there. It works today because `timedSections` (line 587) does list it, so a change cascades down into a fresh `positionedSections`. But that's an invisible chain — if anyone ever restructures those memos, this breaks quietly. Just add `laneField`.

packages/raystack/components/data-view/utils/index.tsx
```diff
@@ -113,15 +118,18 @@ export function groupData<TData>(
   const groupCountMap = field?.groupCountMap || {};
   const groupedData: GroupedData<TData>[] = [];
 
-  groupMap.forEach((value, key) => {
+  // Section order: the field's declared `groupOrder` first, then undeclared
+  // values in first-seen order, then the empty (null-valued) bucket last.
+  for (const key of orderBucketKeys([...groupMap.keys()], field?.groupOrder)) {
```
Shreyag02, line 123: This changes behaviour for people not using the new feature

The empty bucket (rows with no value) now always sorts **last** — including when the field declares no `groupOrder` at all. Before this PR, `groupData` used `groupMap.forEach`, which kept insertion order, so that bucket stayed wherever it first appeared.
 
Meaning: anyone grouping a `DataView.List` on a nullable field can see their section order move, without opting into anything. No test caught it because the new `group-data.test.ts:77` locks in the new behaviour rather than comparing to the old.

we can add a line in the PR description, since there's no `.changeset/` in this repo to carry it.

packages/raystack/components/data-view/utils/pack-lanes.tsx
```diff
@@ -47,6 +48,61 @@ export function packLanes(
...
+ * hand over an already-ordered list (the timeline passes the sorted row model),
+ * so first-seen *is* the caller's order. Within a bucket the greedy first-fit
+ * of `packLanes` decides sub-lanes, so a value with no overlapping cards
+ * occupies exactly one lane.
+ */
+export function packLanesBySortValue(
+  items: PackSortValueLaneItem[],
+  gapPx: number = DEFAULT_CARD_GAP_PX
+): PackLanesResult {
+  const lanes = new Array<number>(items.length).fill(0);
```
Shreyag02, line 77: allocates the array before the empty-input return. Swap the two lines.

Shreyag02 (changes requested):

apps/www/src/content/docs/components/dataview/index.mdx
```diff
@@ -474,6 +516,8 @@ Bands are labels only in this release: no chevron, no collapsing.
 
 Sort can't move a card horizontally — x is locked to the start date — so it surfaces in exactly one place: `lanePacking="one-per-row"`, where vertical row order follows the active sort (within each section when grouped). With the default `auto` packing, lanes are assigned by dense chronological first-fit and the sort has no visible effect, so hide the Ordering control (`<DataView.DisplayControls hideOrdering />`) or leave `sortable` off the timeline's per-view `fields` unless you use `one-per-row`.
 
+`lanePacking="one-per-field"` is the other place sort reaches, and there it does more than reorder: the sorted-by field *defines* the lanes, so changing the sort field rebuilds them (see [Lane packing](#lane-packing)). Leave the Ordering control visible for that mode.
+
```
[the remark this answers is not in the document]
rsbh: Fixed in 5a20a0ed — you were right that the section contradicted itself; the paragraph I appended below said the opposite of the lead sentence. Rewritten around "two of the three packing modes": `one-per-row` reorders rows, `one-per-sort-value` rebuilds the lanes, and only `auto` is the mode where hiding the Ordering control makes sense.

packages/raystack/components/data-view/components/timeline.tsx
rsbh: Real bug, fixed in 5a20a0ed. Confirmed your reading in the installed `table-core@8.9.2`: `column.js` builds a deep `accessorFn` when the key contains a dot, so `original["meta.rank"]` is undefined for the exact key the sort used. Now reads `row.getValue(laneField)`, which is what the sort saw, and `getValue` returning undefined for an unknown column (`row.js:27-36`) is what makes the no-column case safe rather than throwing.

Two additions beyond the fix:
- A test with `accessorKey: "meta.rank"` and ranks 1/2/3 pins lanes 0/1/2.
- The silent-collapse class of failure had a second entrance: a sort key matching no field at all still reads undefined on every row. That now warns too ("which matches no field"), with its own test. Worth noting the lanes there are `0,1,0,2` rather than all-zero — one bucket, still sub-laned by date overlap.

packages/raystack/components/data-view/components/timeline.tsx
rsbh: Added in 5a20a0ed. Agreed on the reasoning — the cascade through `timedSections` made it correct but invisible.

packages/raystack/components/data-view/utils/pack-lanes.tsx
rsbh: Swapped in 5a20a0ed — returns `{ lanes: [], laneCount: 0 }` before allocating.

packages/raystack/components/data-view/utils/pack-lanes.tsx
rsbh: Took your option (b) in 5a20a0ed: lanes follow the sort, full stop — that is the deliberate contract of the mode, so the code stayed and the comments were what needed fixing.

Kept the `orderBucketKeys` call, though: with no `order` argument it is still doing the empty-bucket-last half of the rule, which lanes do need. Both comments now say so explicitly — `order-bucket-keys.tsx` no longer claims sections and lanes "can never disagree", and `packLanesBySortValue` states that a field which is both grouped and sorted can rank its sections by `groupOrder` and its lanes by the sort, by design.

On (a) and the demo `rank` field: that was the earlier design and it was dropped on purpose — `laneField`/`laneOrder` existed for it and both were removed so the sort is the single vocabulary. So the demo keeps `rank`.

packages/raystack/components/data-view/utils/index.tsx
rsbh: Correct, and it is intentional — the empty bucket now sorts last whether or not `groupOrder` is declared, so section order can move for anyone grouping a `DataView.List` on a nullable field without opting in.

It is already in the PR description (the "Also: declared section order" paragraph, plus the note that this is the one non-additive change). I have expanded it to name the `DataView.List` impact directly rather than leaving it implied.

On the test: `group-data.test.ts` pins both paths deliberately — "puts rows with no value in the last section" for the declared-order case and "puts rows with no value last without a declared order too" for the case you flagged. You are right that neither compares against the old order, since the old order is what is being changed; the second test is there to make the change explicit rather than incidental.

packages/raystack/components/data-view/utils/pack-lanes.tsx
```diff
@@ -47,6 +48,65 @@ export function packLanes(
...
+    if (!bucket) {
+      bucket = [];
+      buckets.set(key, bucket);
+    }
+    bucket.push(index);
+  }
+
+  let laneCount = 0;
+  for (const key of orderBucketKeys([...buckets.keys()])) {
+    const bucket = buckets.get(key) as number[];
```
Shreyag02, line 94: A bucket of one still pays for `orderByX` (a sort plus two typed arrays), and with a high-cardinality sort field that's most of the buckets — and the Ordering control lets a user pick one at runtime.

50k rows, all-distinct keys: **46 ms → 20 ms**. Identical lane output, no change in the few-bucket case.

```suggestion
    const bucket = buckets.get(key) as number[];
    // A bucket of one can't collide with itself.
    if (bucket.length === 1) {
      lanes[bucket[0]] = laneCount++;
      continue;
    }
```

packages/raystack/components/data-view/components/timeline.tsx
```diff
@@ -535,7 +563,24 @@ export function DataViewTimeline<TData>({
...
+        // else (an object, an array) shares the no-value lane rather than
+        // collapsing into one "[object Object]" bucket.
+        let laneKey: string | null = null;
+        if (sortValueLanes) {
+          // Read through the row, not `original`: TanStack treats a dotted
+          // `accessorKey` as a path, so `original['meta.rank']` is undefined
+          // for the very key it sorted by — every row would look valueless and
+          // pile onto one lane. `getValue` yields exactly what the sort saw
+          // (undefined rather than a throw if the key names no column).
+          const value = row.getValue(laneField as string);
```
Shreyag02: Follow-on from the `row.getValue()` change, and the one thing I'd want in before this lands.

`getValue` re-resolves the column on every call, and `table.getColumn()` logs a dev error when it misses — and TanStack doesn't cache the miss (`row.js` returns before writing `_valuesCache`). So the sort-key-matches-no-field case now emits one `[Table] Column with id '…' does not exist.` **per row, per recompute**. Your new test prints 6 of them for 4 rows; at a few thousand rows it buries the warning you added just above.

Resolving the column once fixes both:

```tsx
const laneColumn = useMemo(
  () => (sortValueLanes && table ? table.getColumn(laneField as string) : undefined),
  [sortValueLanes, laneField, table]
);

// in the row loop
const value = laneColumn ? row.getValue(laneField as string) : undefined;
```

The `useEffect` can then test `laneColumn` rather than calling `getColumn` a second time. Applied locally: errors go 6 → 2 and stop scaling with row count; 228/228 data-view tests still pass.

packages/raystack/components/data-view/__tests__/timeline.test.tsx
```diff
@@ -303,6 +303,10 @@ type Order = {
   start: string | null;
   end: string | null;
   team?: string;
+  // biome-ignore lint/suspicious/noExplicitAny: one-per-sort-value takes any value
+  priority?: any;
```
Shreyag02: Tiny one: `unknown` works here — `priority` is only ever written into test data, never read in a typed position — so the suppression can go with it.

```suggestion
  // one-per-sort-value reads whatever the sorted field holds.
  priority?: unknown;
```

`tsc --noEmit` stays identical to base with this.

Shreyag02 (approved):
Approving. All five of my earlier comments are addressed, and I re-derived the packing logic rather than re-reading it — I couldn't construct a case where a lane holds two values or where two cards in one lane overlap. The property tests are the right shape for this.

Verified locally at 5a20a0e: 228/228 data-view tests, `tsc --noEmit` produces the identical error set to base (zero new), `build:apsara` green, and a trial merge with current `main` is clean.

Three small non-blocking suggestions inline. The first is a side effect of the `row.getValue()` change my earlier comment prompted, so it's worth catching before this lands. I applied all three locally — TS clean, no new lint warnings, tests still pass.

One follow-up worth a separate issue rather than this PR: now that lanes follow the sort and the demo keeps a synthetic `rank`, having `TimelineCardContext` carry the lane value would let a card label its own lane without re-deriving it from `row.original[sortField]` — the same dotted-key trap we just fixed. `laneKey` is already on the item.

Two notes for the docs, whenever: server mode takes lane order from row order, so a backend that ignores `sort` gives arbitrary lane order with no warning — worth a sentence in the Ordering section. And the description says 44 new tests / 2681 in the package; I count 41 new `it()` blocks and 2677, probably just because the branch is behind `main`.

rohanchkrabrty (approved):

PR #819, opened 2026-05-21
fix(calendar): typing, controlled-value, slotProps API, tz-aware day keys

Shreyag02:
## Description

  Coordinated overhaul of `Calendar`, `DatePicker`, and `RangePicker` to fix functional drift across the three surfaces.
 
  **Highlights:**
   Description

  Coordinated overhaul of Calendar, DatePicker, and RangePicker to fix functional drift across the three surfaces.

  Highlights:

  - 18 P0/P1 bugs fixed, including a P0 hotfix for the "keyboard input works only once" crash on DatePicker (dayjs.extend was missing for `isSameOrAfter / isSameOrBefore`— every keystroke threw).
  - New `slotProps API` on both pickers ({ input?, calendar?, popover? } for DatePicker; { startInput?, endInput?, calendar?, popover? } for RangePicker) replacing the inconsistent inputProps / inputsProps / calendarProps / popoverProps. Legacy props kept as @deprecated aliases for one release.
  - DatePicker is now reactive to value, has a defaultValue prop, separates view-month from selected date, strict-parses typed input against dateFormat, and supports a true unselected state (no more silent new Date() fallback).
  - RangePicker state machine rewritten, {today, today} default removed, controlled value syncs the visible month including on clear, adopts the new usePickerPopover hook for outside-click + year/month dropdown carve-out.
  - disabled actually disables the picker. The trailing calendar icon renders as a sibling <div> to the <input>, so its clicks bubble to Popover.Trigger and used to open the calendar even when the input was disabled. Both pickers now gate the popover off the merged input's disabled. RangePicker treats either input being disabled as fully disabled — partial-disable was leaky because the shared range state machine rewrites both from and to regardless of which input was clicked. For "fix one side, pick the other", constrain the calendar via calendarProps instead.
  - dateInfo icons no longer render as solid blobs when their day is selected — the .dayInfo svg { fill: emphasis } rule was overriding fill="none" on stroke-based icons (lucide). color alone now carries the selected style via currentColor for both stroke- and fill-based icon libraries.
  - Calendar dateKey is now tz-aware so tooltipMessages / dateInfo lookups match the rendered day; onDropdownOpen no longer re-fires every render.
[...]

PR #890, opened 2026-08-24
feat: RFC calendar rewrite (CalendarPreview)

Shreyag02:
Proposes replacing `Calendar`, `DatePicker` and `RangePicker` with a single subcomposed root that owns every piece of state explicitly — selection, view month, popover open, scale, validity — and exposes every surface as a dot-notation part.

## Description

Adds `docs/rfcs/005-calendar-preview.md`. **Docs only — no component code in this PR.**

The calendar family is the only part of Apsara that never adopted the composition contract. The RFC argues every recurring bug is downstream of that, and proposes `CalendarPreview`: one export, dot-notation parts, `value`/`onValueChange` + `open`/`onOpenChange`, react-day-picker isolated behind one file.

Five problems account for it — each a question about who owns what, not about rendering:

| Problem today | Proposed fix |
|---|---|
| **Open state is private**, so dismissal can't go to Base UI — 185 lines of bespoke popover code, and `captionLayout='dropdown'` is off by default because the `Select`s it mounts loop on unmount | `open`/`defaultOpen`/`onOpenChange` on the root; `use-picker-popover.ts` deleted |
| **No composition contract** — three flat exports, zero sub-parts; `slotProps`, `children`-as-function and `onErrorChange` appear nowhere else in the library | One root, every surface a part, zero `slotProps` |
| **RDP's prop union *is* the public API** — it leaks, `props.ts` is a hand-maintained mirror that has drifted, and spread-last is unsatisfiable in the pickers | RDP reaches only `.Grid`, driven by derived props |
| **`Date` identity churn is load-bearing** — three `biome-ignore`s plus a fourth effect left unguarded, and `dayjs.extend()` in four modules in an import order enforced by a comment (the 0.49.0 P0) | `dayKey()`/`epoch()` internally; one `date-adapter.ts` |
| **Coarser-than-day selection has no home** — a `Date` can't say whether it means "August 2026" or "1 August 2026", and cells and chips render it with no calendar mounted | `ScaleValue` — the value carries its own scale |

### The new capability

Selection at scales coarser than a day — month, quarter, half-year, year — with the scale carried by the value rather than inferred from a prop:

```ts
interface ScaleValue {
  date: string;   // 'YYYY-MM-DD', timeless
  scale: 'day' | 'month' | 'quarter' | 'halfYear' | 'year';
}
```

`scale` is the view that is open; `value.scale` is what the committed value means. They differ while the user browses, a prop can't say what a stored date meant after a reload, and one object makes an inconsistent pair unrepresentable. `trailingValue` picks which edge of the period is emitted (a start field emits 1 Aug, an end field 31 Aug — and February 2028 trailing is `2028-02-29`), and availability is judged on **the date a period would produce**, so one period can be selectable in a start field and disabled in an end field. The period maths lands as pure functions in `lib/`, tested before any UI.

### Behavioural breaks a codemod cannot find

- **Ranges emit only on completion.** `to` is no longer nullable; anyone relying on the first-click event loses it
- **Bounds limit selection, not navigation.** `minDate`/`maxDate` disable out-of-range periods; the view still moves. A deliberate change from `startMonth`/`endMonth`
- **A read-only endpoint needs a value.** `readOnly` on one `.Input` replaces `lock` gating the whole picker
- **`mode="multiple"` is removed.** `Calendar` accepts `mode="multiple"` with `selected: Date[]` today — shipped, documented public API — and v1 has no equivalent. This is a **removal of shipped surface, not an unbuilt feature**; the RFC gates it on a consumer search before the deletion phase. Flagging it here so it isn't a surprise at release-notes time

Every call site also becomes a composition, because the pre-composed recipes were cut.

### Findings turned up while writing it

Each is cited in Appendix B — to types, named effects, source comments and changelog headings rather than line numbers, so the citations survive the next merge from `main`.

| Where | Finding |
|---|---|
| `date-picker.tsx` | Unguarded effect loops on an inline `defaultMonth`; its three siblings were hardened, this one missed |
| `range-picker.tsx` | `RangePickerProps`/`RangePickerSlotProps` unexported and in no barrel — consumers can't type a wrapper, yet the docs render its type table |
| `props.ts` | Types `slotProps.calendar` as the full docs `CalendarProps`, including `mode`/`selected`/`onSelect`/`footer`, none of which the real slot type accepts. Five of the six deprecated props are missing from it entirely |
| `CHANGELOG.md` | Claims a `DataView` `filterProps.calendar` slot that was never built — `data-view.types.tsx` has only `{ select? }` |
| `filter-chip.tsx` | Shallow `slotProps.input` merge silently drops the chip's container class; two CSS rules reach at `Input` hashed classes that appear nowhere in the repo and are dead code |
| `calendar.module.css` | Three `Todo: var does not exist` markers, a hardcoded `max-height: 260px`, eight `var(--rs-space-10, 40px)` fallbacks — and no interactive `playground` in `demo.ts`, which 54 of 69 components have |
| lockfile | date-fns and `@date-fns/tz` already ship unconditionally via RDP *and* are peered by Base UI; dayjs is a second date implementation we pay for twice, so the adapter is built on date-fns |
| Base UI 1.6/1.7 | No date component, but `internals/temporal` plus date-fns and Luxon adapters exist — `date-adapter.ts` is shaped to that surface, so adopting theirs later is a one-file swap |

Breaking, no shim: the new component ships alongside, the old family is removed one release later. Seven phases; the three migrated components reach parity **before** the scale surface is built, so the scale model ships second even though it is specified first. Four open items for review — the `scales` discriminator typing, `.Picker`'s name, whether `selection='multiple'` must ship before the deletion, and how the break gets announced given there is no changesets setup.

## Type of Change

- [x] Documentation update
- [ ] Improvement (Improvements to existing code)
- [ ] Other (please specify)

Docs-only PR, so nothing here is breaking. The change it *proposes* is breaking and includes a removal of shipped public API (`mode="multiple"`) — noted above so the label isn't a surprise later.

## How Has This Been Tested?

No code, so nothing to run. The RFC's claims were verified instead, and re-verified against the current branch:

| Claim | How |
|---|---|
| Every citation in Appendix B | Re-checked against the working tree — line counts, the three `biome-ignore`s, the unguarded effect, the six refs and three suppression branches in `use-picker-popover.ts`, the 23 `data-slot`s, the dead `FilterChip` CSS, the CSS token violations, the 54-of-69 playground count, and the two tests behind each of the dayjs P0 and the partial-disable gate |
| Dependency versions | npm registry + `pnpm-lock.yaml`, checked today: manifest, lockfile and latest all match the RFC's table |
| RDP v10 union unchanged | Diffed the 9.6.7 and 10.0.1 tarballs — `types/selection.d.ts` gained JSDoc only, so the upgrade does not by itself fix the union. v10 also drops `date-fns-jalali` and goes from 16 `@deprecated` props to zero, none of which we reference, and keeps the `DayButton`/`Weekday`/`MonthGrid` slots the RFC binds to |
| Base UI has no date primitive | Enumerated its exports at 1.6.0 and 1.7.0 — `./internals/temporal`, `./internals/temporal-adapter-date-fns`, `./internals/temporal-adapter-luxon`, no top-level date component. 1.7.0 pins `@base-ui/utils` at exactly 0.3.2 |

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas — every claim backed in Appendix B
- [x] I have made corresponding changes to the documentation (.mdx files) — this PR *is* the doc; component `.mdx` lands with the implementation
- [ ] I have added tests that prove my fix is effective or that my feature works — N/A, no code; acceptance criteria are in the RFC's Testing section

N/A — no visual change.

## Related Issues

| | |
|---|---|
| Prior art | #819 — the coordinated overhaul this RFC argues was necessary but not sufficient |
| Lands with the rewrite | `FilterChip` rewrite, `DataView` `filterProps.calendar` slot, `filter-operations` → `date-adapter.ts` |
| No longer independent | `react-day-picker` → 10.0.1 should land **before phase 2** — not because it fixes the union (it doesn't), but because `.Day`/`.Weekday` bind to RDP's component-override slots and `.Grid` runs with `hideNavigation`, so those names must be pinned to whichever major ships |
| Independent | `@base-ui/react` → `~1.7.0` (moves `@base-ui/utils` to 0.3.2 with it) |


rohanchkrabrty (commented):
@Shreyag02  can you make the RFC a bit more concise? Currently it feels bloated and hard to understand.

Shreyag02:
Thanks — trimmed it down: 694 → 548 lines, 6.3k → 5.3k words.

- Background is now bold-lead paragraphs matching RFC 004's style, and the Scorecard table folded into a single paragraph.
- Design material is out: Design Blockers, phase 0, and every Figma reference. The one engineering residue (CSS uses `--rs-*` tokens only) already lives in the testing checklist.
- Dependencies, the `data-slot` mapping, Conventions, and Alternatives lost restatement rather than claims.
- Citations no longer use line numbers — they point at types, named effects, source comments, and changelog version headings, so they don't rot on the next merge from main.

Also fixed the stale citations and slot-table gaps along the way; every reference is now verified against the branch.

The API surface, migration map, implementation plan, and open items are unchanged.

docs/rfcs/005-calendar-preview.md
rohanchkrabrty, line 1: ## 1. Rename `granularity` to `scale`

Matches `TimelineScale` in `DataView` — one word across the library.

```ts
type Scale = 'day' | 'month' | 'quarter' | 'halfYear' | 'year';

scale?: Scale;                      // controlled active scale
defaultScale?: Scale;
onScaleChange?: (scale: Scale) => void;
scales?: Scale | Scale[];           // supported scales; single value hides the switcher
```

## 2. The value must carry its own scale

`value?: Date | null` makes "August 2026" and "1 August 2026" the same value. It needs to be a date/scale pair:

```ts
interface DateValue {
  date: string;   // 'YYYY-MM-DD', timeless
  scale: Scale;
}

value?: DateValue | null;
defaultValue?: DateValue | null;
onValueChange?: (value: DateValue | null, details: ChangeDetails) => void;
```

It can't come from the `scale` prop — that's the view that's open, `value.scale` is what the committed value means. They differ the whole time the user is browsing another scale; after a reload the prop can't say what a stored date meant; and consumers render the value with no calendar mounted (`FilterChip` text, DataView filters), where there's no prop to read. Separate `value` and `scale` props can also be set inconsistently — one object makes that unrepresentable.

Use `'YYYY-MM-DD'` strings, not `Date`. `new Date('2026-08-01')` is UTC midnight and renders as 31 July in negative offsets, and `Date` identity churn is what the `biome-ignore`s and the `setViewMonth` loop already are. Convert once, inside the grid, where react-day-picker needs it.

## 3. Add `trailingValue`

```ts
trailingValue?: boolean;   // default false
```

"August 2026" stores `2026-08-01` on a start field and `2026-08-31` on an end field — same for quarter, half-year and year. It decides which date we emit, so it belongs in the value contract and can't be added later without changing what call sites receive. Month-end must be calendar-correct: Feb 2028 trailing is `2028-02-29`.

Unrelated to `lock='from' | 'to'`, which is about range endpoints.

## 4. Draft state

The state table is missing the draft. Without it, switching scale either emits — writing on what the user sees as a preview — or does nothing until Apply.

| State | Source | Emitted |
|---|---|---|
| `value` | controlled prop, else internal | on commit |
| `draft` | scale switch or typing; null when clean | never |
| `scale` | controlled, else follows `value` or typed input | `onScaleChange` |
| `viewMonth` | controlled, else `draft ?? value ?? today` | `onMonthChange` |

Switching scale or typing sets the draft and moves the view without emitting. Clicking a cell or pressing Enter commits. Escape drops the draft and restores the input from `value`.

## 5. Define the scale conversion rule

Currently unspecified. One rule covers every direction:

> Take the anchor date. Find the period of the target scale containing it. Emit that period's start when leading, its end when trailing.

```ts
convertScale(value, to, trailing) {
  return { date: anchorOf(periodOf(value.date, to), trailing), scale: to };
}
```

Availability is the same rule — a cell is available when the date it would produce falls inside `[min, max]`. With a start of 15 Jul 2026 an end picker disables H1 2026 (ends 30 Jun) but keeps Q3 2026 (ends 30 Sep). `isDateUnavailable(date: Date)` as typed can't express that.

Build this as pure functions in `lib/`, tested before any UI: leap years, month-end snapping, period boundaries, round trips on both edges.

## 6. `onValueChange` needs details

```ts
interface ChangeDetails {
  reason: 'select' | 'input' | 'clear' | 'scale';
  period: { start: string; end: string };
  toDate: () => Date;
}
```

Without `period` every consumer re-derives the span by hand and gets month-end wrong. `period` doesn't depend on `trailingValue`, so a consumer that ignores the edge still gets the right span.

## 7. One view part per scale, not one `.MonthGrid`

A single part switched by a root prop can't render only the quarter view, and the four views need different layouts — 3 / 4 / 2 / 1 columns, and the year view has no year headings.

`DataView` already does this: `.List`, `.Timeline` and `.Custom` are sibling view parts on one root, each self-gating with

```ts
const isActive = !name || activeView === undefined || activeView === name;
```

Mount one alone and it always renders; mount several and the switcher picks. Use `.Days`, `.Months`, `.Quarters`, `.HalfYears`, `.Years`.

`.MonthGrid` needs renaming anyway — it collides with react-day-picker's own `MonthGrid` slot and with today's `data-slot="calendar-month-grid"`, which means the day table.

## 8. `.Nav` → `.Header`, with subparts

`.Nav` only applies to the day scale, and `align` as its only prop makes a custom header impossible.

```tsx
<Calendar.Header>
  <Calendar.PrevMonth />
  <Calendar.Title />
  <Calendar.Reset />
  <Calendar.NextMonth />
</Calendar.Header>
```

`.Header` composes these by default, so the common case doesn't get longer. It sits outside the grid, with react-day-picker on `hideNavigation`.

The third button is a reset — return the view to the anchor month, shown only once the view has drifted. Different from a "Today" jump in the footer; both can exist.

## 9. `.Input` mounts anywhere under the root

Not just inside `.Trigger`. The design puts it inside the popup, above the scale switcher. Use a `field` prop rather than `.RangeInput` with `startProps` / `endProps`:

```tsx
<Calendar.Input field="start" />
<Calendar.Input field="end" />
```

`startProps` / `endProps` is `slotProps` renamed, against the stated zero-`slotProps` goal, and it stops you putting anything between the two inputs or rendering only one. Same for `.Grid`'s `dayProps` — that should be a `.Day` part on react-day-picker's `DayButton` slot.

## 10. View props go on views

**Move down:** `numberOfMonths` → `.Days`; `fixedWeeks`, `showOutsideDays`, `showWeekNumber`, `weekStartsOn`, `modifiers` → `.Grid`; `onValidityChange` → `.Input`.

**Add to root:** `yearRange` (which years the period views cover), `today` (deterministic tests and playground), `clearable`.

`open` / `defaultOpen` / `onOpenChange` stay on the root, matching `Dialog` / `Popover` / `Menu` — inert when no `.Trigger` / `.Content` is mounted, which is what makes the inline calendar the same component rather than a mode.

## Proposed structure — additions to the part tree

`.Trigger`, `.Content`, `.Grid` and `.Footer` stay as proposed, so there's no separate picker component. Everything marked below is an addition or a rename on top of that tree.

```tsx
<Calendar>                              {/* value · draft · scale · view · bounds */}
  <Calendar.Trigger />                  {/* omit Trigger + Content for inline */}
  <Calendar.Content>
    <Calendar.Label />                  {/* new */}
    <Calendar.Input field="end" />      {/* moves — anywhere under the root */}
    <Calendar.Scales />                 {/* renamed from .GranularityTabs */}
    <Calendar.Separator />              {/* new */}

    <Calendar.Panel>                    {/* new — no children renders the views below */}
      <Calendar.Days>                   {/* new */}
        <Calendar.Header>               {/* renamed from .Nav */}
          <Calendar.PrevMonth /> <Calendar.Title />      {/* new */}
          <Calendar.Reset />     <Calendar.NextMonth />  {/* new */}
        </Calendar.Header>
        <Calendar.Grid>                 {/* react-day-picker, only here */}
          <Calendar.Day />              {/* new — → DayButton slot */}
          <Calendar.Weekday />          {/* new — → Weekday slot */}
        </Calendar.Grid>
      </Calendar.Days>

      <Calendar.Months />               {/* these four replace .MonthGrid */}
      <Calendar.Quarters />
      <Calendar.HalfYears />
      <Calendar.Years />
    </Calendar.Panel>

    <Calendar.Footer>
      <Calendar.Clear /> <Calendar.Today />   {/* new */}
    </Calendar.Footer>
  </Calendar.Content>
</Calendar>
```

`.Presets`, `.Preset` and `.TimeField` stay as proposed and are left out above only to keep the tree short. `.Apply` / `.Cancel` go with `commit`.

Every part renders its own default with no children, so composition is opt-in depth. The tree above collapses to:

```tsx
<Calendar value={target} onValueChange={setTarget} min={start} trailingValue>
  <Calendar.Trigger />
  <Calendar.Content />
</Calendar>
```

Inline, day grid only:

```tsx
<Calendar scales="day"><Calendar.Panel /></Calendar>
```

Quarter-only — no day grid mounted, no switcher rendered:

```tsx
<Calendar scales="quarter" trailingValue>
  <Calendar.Trigger />
  <Calendar.Content><Calendar.Quarters /></Calendar.Content>
</Calendar>
```

`.Scales` renders every scale in `scales` on its own. `.Scale` is only for relabelling or reordering:

```tsx
<Calendar.Scales>
  <Calendar.Scale value="quarter">Quarterly</Calendar.Scale>
</Calendar.Scales>
```

**Root prop changes**

Everything else the RFC proposes stays as it is — `format`, `timeZone`, `open` / `defaultOpen` / `onOpenChange`, `disabled`, `readOnly`, `selection`, `lock`. These are the props that change:

```ts
// changed shape
value?: DateValue | null;
defaultValue?: DateValue | null;
onValueChange?: (value: DateValue | null, details: ChangeDetails) => void;

// changed type — 'YYYY-MM-DD' strings, was Date
minDate?: string;  maxDate?: string;
isDateUnavailable?: (date: string) => boolean;
month?: string;  defaultMonth?: string;
onMonthChange?: (month: string) => void;

// renamed
scales?: Scale | Scale[];                 // was granularities; default all five
scale?: Scale;                            // was granularity
defaultScale?: Scale;
onScaleChange?: (scale: Scale) => void;

// added
trailingValue?: boolean;                  // default false
yearRange?: { from: number; to: number };
today?: string;
clearable?: boolean;

// removed from the root
// commit           — see #4
// weekStartsOn     — moves to .Grid
// onValidityChange — moves to .Input
```

`onOpenChange` should forward Base UI's typed event details rather than redeclaring `{ reason?: string }`.

**Part props**

| Part | Props |
|---|---|
| `.Trigger` | `render`, `Popover.Trigger` props |
| `.Content` | `side`, `align`, `Popover.Content` props |
| `.Days` | `numberOfMonths` |
| `.Grid` | `fixedWeeks`, `showOutsideDays`, `showWeekNumber`, `weekStartsOn`, `modifiers`, `components` |
| `.Input` | `field`, `placeholder` (a `DateValue`), `onValidityChange`, `Input` props |
| `.Scale` | `value`, `render` |
| view parts | `name`, `columns`, `render` |
| every part | `render`, `className`, `ref`, `data-slot`, `children ?? default` |

Every part should take `render` — the house pattern in `composition.md`, currently on 4 of 13 — and children should override context-computed content the way `Tour.Title` does, so `<Calendar.Title>Q3 2026</Calendar.Title>` works. That also removes the need for the pre-composed recipes. Please add a `useCalendar()` hook like `useTour()` so consumers can build parts we didn't ship.

## Smaller points

- Add state `data-*` attributes alongside `data-slot`: `data-selected`, `data-draft`, `data-unavailable`, `data-today`, `data-outside`, `data-scale`. Slots say what an element is, not what state it's in, and custom cells need something to style against.
- The `DataView` `filterProps.calendar` gap is real but should be separate from this.

rohanchkrabrty (changes requested):

docs/rfcs/005-calendar-preview.md
Shreyag02: Thank you for the detailed review — it's reshaped the RFC, and I've rewritten rather than patched. Where each point landed:

1. **`scale` — agreed**, your type as written, `halfYear` in code with "Half-year" as the label. Added a docs note that the value set differs from `DataView`'s `TimelineScale`, which has `week` and stops at quarter.

2. **Value shape — adopted for the new input; `Date` stays on the migrated three.** The scale input takes `{ date, scale }` with `toDate()` as you proposed. Keeping `Date` on `Calendar` / `DatePicker` / `RangePicker` does leave your concern unaddressed for those — UTC-midnight rendering and identity churn stay in their public surface — but ISO strings break FilterChip and DataTable's filter types and the migration stops being mechanical. Logged as an accepted cost with a follow-up for a future major. One consequence: `scales` defaults to `'day'`, since that's what discriminates the two shapes.

3. **`trailingValue` — adopted.** Root prop, default false; start emits the period's first day, end its last, month-end correct.

4. **Draft state — adopted.** A scale switch moves the view and sets a draft; nothing emits until a pick or Enter, and Escape drops it.

5. **Conversion and availability — adopted**, as tested pure functions in `lib/`. Availability tests the date a period would produce, so a period can be selectable for a start and disabled for an end — which I read as the point of your 15 July example.

6. **`toDate()` in; `period` still owed.** In the rewrite.

7. **View parts — adopted.** `.Days` / `.Months` / `.Quarters` / `.HalfYears` / `.Years`, self-gating like DataView's. `.MonthGrid` and its RDP slot collision both gone.

8. **`.Header` split — adopted, two things I read differently.** The third button is a value reset rather than a view reset: it restores `defaultDate` and renders only when that's set and the selection differs. And the caption month/year dropdown stays on the standalone calendar — our own two-column scroller, no `Select`, so the unmount loop can't return. Happy to walk the frames if that reading looks off.

9. **`.Input` placement — same problem, different solution.** Start and end are independent single-value roots, each with its own scale, since the design has `Aug 1st → Q3 2026`. No `startProps` / `endProps`; `lock` becomes per-input `readOnly`.

10. **View props and part API — adopted in full**, including `useCalendar()`, which matters more now that the pre-composed recipes are cut entirely.

Beyond your review: `commit` / `.Apply` / `.Cancel` gone (commit on pick, Enter, blur, outside click); `.Presets` / `.Preset` / `.TimeField` out of v1, no design for any; `format?: string` replaced by a formatter function. Ranges emit only on completion, so `to` is no longer nullable. `multiple` parked. State `data-*` attributes still outstanding.

One offer: a few of the conventions you pointed at — state attributes, consumer hooks, children-override, sibling view parts — aren't in `composition.md`. Happy to write that section from this review if it'd help.

rohanchkrabrty (approved):

PR #893, opened 2026-09-02
feat(theme): revamped Theme

rohanchkrabrty:
## Summary

- Adds `ThemePreview`, the element-mounted theme from RFC 004: tokens move off `<html>` onto a real element, so the root theme, a nested scope and a portal re-injection are the same component rendering the same attributes — which makes the theme server-renderable, allows more than one provider per page, and fixes scoped themes inside portals.
- Rebuilds the token layer: every `--rs-*` declaration is wrapped in `:where()` so a consumer rule on the stable `.rs-theme` class wins without `!important`; spacing, effects and z-index move onto the theme selector so a scope can change them; adds `--rs-scaling`, a radius factor over a fixed base scale replacing the two hardcoded `data-style` scales, and panel/overlay/backdrop-filter tokens.
- Adds seven independently seedable, controllable and persistable settings — appearance, accent, gray, radius, scaling, panel background, reduced motion — over `useSyncExternalStore` persistence gated on `persistKey`, plus a small pre-hydration script that patches its own parent and a `:has()` colour-scheme rule that writes nothing to `<html>`.
- Adds theme re-injection across the thirteen portalling components, and a shared `radius` cva variant and CSS module; `Image` and `Avatar` migrate onto the five-value scale.
- Publishes `style-no-fonts.css` alongside `style.css`, and adds a theme playground panel plus a docs page with a migration guide.
- The existing `Theme` ships unchanged, so this is additive — no consumer migration is forced by this PR.


packages/raystack/components/theme-preview/store.ts
```diff
@@ -0,0 +1,138 @@
...
+    const next = patch[key];
+    if (next === undefined || settings[key] === next) continue;
+    settings[key] = next;
+    changed = true;
+  }
+  if (!changed) return;
+
+  const entry: StoredEntry = { v: STORAGE_VERSION, settings };
+  try {
+    window.localStorage.setItem(persistKey, JSON.stringify(entry));
```
Shreyag02, line 111: Theme changes get dropped when `localStorage` is unavailable.

For persisted keys, storage is the source of truth. `readRaw` (L41) and this `setItem` both catch-and-return, so the write is lost, `notifyThemeStorage()` on L115 still fires, readers re-read the previous value, and the setting doesn't move.

Reproduced: with `persistKey` set and a throwing `localStorage`, appearance stays `light` on click; without `persistKey` it toggles fine. That covers Safari private mode, sandboxed iframes, and browsers with site data blocked — and the more correct config (with `persistKey`) is the affected one.

Could `writeStoredSettings` return whether it succeeded, so the caller can fall back to in-memory state?

packages/raystack/components/theme-preview/theme-preview.tsx
```diff
@@ -0,0 +1,487 @@
...
+        // does not know which keys a page controls stays usable.
+        if (valueRef.current?.[key] !== undefined) continue;
+        const allowed: readonly string[] = SETTING_VALUES[key];
+        if (!allowed.includes(candidate)) continue;
+        assignSetting(patch, key, candidate);
+      }
+      if (isSettingsEmpty(patch)) return;
+
+      const persisted = persistedKeysRef.current;
+      if (persistKey && persisted.length > 0) {
```
Shreyag02, line 317: Companion to the `store.ts:111` comment — this is where the fallback would live. When the write fails there's no path to `setLocal`, since the loop below skips anything in `persisted`.

Something like `if (!writeStoredSettings(...)) { /* mirror the patch into local */ }` would keep the switcher working when storage is unavailable.

packages/raystack/components/theme-preview/script.ts
```diff
@@ -0,0 +1,81 @@
...
+    `var d=document,e=d.currentScript,` +
+    `p=(e&&e.parentElement)||d.querySelector(${inlineJson(fallbackSelector)});` +
+    `if(!p)return;` +
+    `var m=${inlineJson(map)},g=${inlineJson(GRAY_PAIRING)},s={};` +
+    `try{var r=localStorage.getItem(${inlineJson(persistKey)});` +
+    `if(r){var o=JSON.parse(r);` +
+    `if(o&&typeof o==="object"&&typeof o.v==="number"&&o.v<=${STORAGE_VERSION}` +
+    `&&o.settings&&typeof o.settings==="object")s=o.settings}}catch(t){}` +
+    `for(var i=0;i<m.length;i++){` +
+    `var k=m[i][0],a=m[i][1],v=s[k];` +
```
Shreyag02, line 69: `v` is sourced only from storage, so an unstored key is skipped at L72 and the server's `data-theme="light"` survives to first paint. Since `DEFAULT_SETTINGS.appearance` is `system`, that's the first visit for most users.

Verified: `renderToStaticMarkup(<ThemePreview persistKey="app">)` with OS dark and empty storage emits `data-theme="light"`, script present but with nothing to act on.

Passing the effective settings into `createThemeScript` and falling back — `var v = (k in s) ? s[k] : d[k]` — would let `system` resolve through `matchMedia` on a first visit. Worth also emitting the script when appearance is uncontrolled `system` even without a `persistKey`.

packages/raystack/styles/theme.css
```diff
@@ -0,0 +1,182 @@
...
+  --rs-radius-factor: 1;
+  --rs-radius-pill: 0px;
+}
+:where([data-radius="large"]) {
+  --rs-radius-factor: 1.5;
+  --rs-radius-pill: 0px;
+}
+:where([data-radius="full"]) {
+  --rs-radius-factor: 1.5;
+  --rs-radius-pill: 9999px;
```
Shreyag02, line 123: `--rs-radius-pill: 9999px` feeds 89 `max(var(--rs-radius-pill, 0px), …)` sites, which includes every large panel.

Measured in Chromium against the built `style.css`: a dialog inside `data-radius="full"` computes `border-radius: 9999px`. Browsers clamp radius to half the box, so it renders as a stadium — same for dropdowns, the command palette, code blocks, data tables, sidebars and calendars.

Capping panel surfaces (`min(var(--rs-radius-pill), 16px)`, or a separate `--rs-radius-panel-max`) while keeping the raw pill for buttons, badges, chips and inputs would fix it.

packages/raystack/components/avatar/avatar.tsx
```diff
@@ -125,7 +125,7 @@ const avatar = cva(styles.avatar, {
   ],
   defaultVariants: {
     size: 3,
-    radius: 'small',
+    radius: 'medium',
```
Shreyag02, line 128: This default means Avatar no longer follows the theme radius. `.radius-medium.radius-medium` is (0,2,0) and always applied, so `.avatar-size-N` at (0,1,0) can never take effect.

Measured on a size-8 avatar:

| theme radius | with the default class | without it |
|---|---|---|
| `full` | 8px | 9999px |
| `large` | 8px | 12px |

So "renders exactly as the old default did" holds only at factor 1. Every other component (Button, IconButton, Badge, Callout, Chip, Input, TextArea) sets no radius default and does respond to the theme.

Dropping `radius` from `defaultVariants` fixes both — one line. It also makes the 13 `max(var(--rs-radius-pill, …))` declarations in `avatar.module.css` reachable again.

apps/www/src/content/docs/theme/preview/index.mdx
```diff
@@ -0,0 +1,430 @@
...
+<ThemePreview persistKey="app-theme" persist={["appearance"]} />
+```
+
+A namespace is one `localStorage` entry holding one JSON object alongside a schema version. A write **merges**: it applies only the settings its `persist` covers and leaves every other field intact, including fields owned by a theme with a different `persist` on the same namespace. A missing or unparseable entry falls back to the seeded defaults and is overwritten on the next write; a field outside its union is discarded individually.
+
+Sharing a `persistKey` is a supported feature, not a hazard — it is exactly what a documentation page wants when several live examples should switch appearance together. Themes on one namespace stay in step within a document, and across tabs through the `storage` event.
+
+## Server rendering
+
+Accent, gray, radius, scaling and panel background are ordinary props, so React server-renders them correctly on the first byte. Only appearance can differ between the server and the browser, and only when it is uncontrolled and persisted — which is the one case the inline script covers.
```
Shreyag02, line 123: Small correction to go with the `script.ts` fix: the inline script covers the case where a value is **stored**, not every uncontrolled-and-persisted case. An unstored `system` appearance still falls through to the server's `light`, which is the common first visit.

packages/raystack/components/dialog/dialog.module.css
```diff
@@ -12,8 +12,10 @@
 }
 
 .dialogContent {
-  background-color: var(--rs-color-background-base-primary);
-  border-radius: var(--rs-radius-2);
+  background-color: var(--rs-color-panel);
+  backdrop-filter: var(--rs-panel-backdrop-filter);
```
Shreyag02, line 16: `backdrop-filter` has no `-webkit-` fallback, and `rollup.config.mjs` runs postcss with `[postcssImport()]` only, so nothing adds one.

Confirmed in the built CSS: `backdrop-filter: blur(64px)` resolves, `-webkit-backdrop-filter` is empty. On Safari ≤17 and older iOS that means a translucent panel with no blur — `background-color` is already `oklab(… / 0.8)`, so page content reads through the text.

Adding autoprefixer to the postcss chain fixes all 12 sites at once: `combobox:11`, `command:8`, `command:170`, `dialog:16`, `drawer:59`, `menu:14`, `popover:15`, `preview-card:21`, `select:12`, `toast:57`, `tooltip:12`, `tour:87`.

packages/raystack/components/theme-preview/theme-preview.tsx
```diff
@@ -0,0 +1,487 @@
...
+  const context = useMemo<ThemeContextValue>(
+    () => ({ ...handle, isRoot: isRootTheme }),
+    [handle, isRootTheme]
+  );
+
+  // Only a root publishes itself, which is what lets a scope flip the page.
+  const inheritedRoot = useContext(RootThemeContext);
+  const rootHandle = isRootTheme ? handle : (inheritedRoot ?? handle);
+
+  const explicitAppearance = value?.appearance ?? defaultValue?.appearance;
```
Shreyag02, line 364: This reads props rather than resolved state, so a nested scope whose appearance came from storage goes dark without painting.

Verified: `<ThemePreview persistKey="nested" persist={['appearance']}>` with `appearance: 'dark'` stored renders `data-theme="dark"` and **no** `data-rs-background`, while the same scope via `defaultValue` gets `data-rs-background=""`. That leaves dark foreground tokens on a transparent background.

Comparing `resolved.appearance` against the parent's would cover it — or a docs note that persisted scopes should pass `hasBackground` explicitly.

packages/raystack/components/theme-preview/portal.ts
```diff
@@ -0,0 +1,44 @@
+'use client';
+
+import { type RefObject, useMemo } from 'react';
+
+import { useThemeContextOrNull } from './context';
+import { settingsToAttributes, THEME_CLASS } from './settings';
+
+/** Where a portalling component puts its content. */
+export type PortalContainer =
```
Shreyag02, line 9: `PortalContainer` is now the type of a public `container` prop on 13 components, but it isn't exported from `theme-preview/index.tsx` or the package root. The emitted `.d.ts` imports it from `'../theme-preview/portal'`, which isn't in the `exports` map either, so consumers have no way to type a container variable.

packages/raystack/components/theme-preview/index.tsx
```diff
@@ -0,0 +1,46 @@
+export {
+  type ThemeContextValue,
+  type ThemeHandle,
+  type UseThemePreviewReturn,
+  useThemePreview
+} from './context';
+export { type ThemeInjectionProps, useThemeInjection } from './portal';
```
Shreyag02, line 7: Could `PortalContainer` be added here alongside `ThemeInjectionProps`?

Also worth re-exporting `createThemeScript` / `ThemeScriptParams` — those would let consumers put the pre-hydration script in `<head>`, which is the usual Next.js answer to the first-paint issue.

packages/raystack/index.tsx
```diff
@@ -144,6 +144,44 @@ export { Table } from './components/table';
 export { Tabs } from './components/tabs';
 export { Text } from './components/text';
 export { TextArea } from './components/text-area';
+export {
+  ACCENT_COLORS,
+  type AccentColor,
+  APPEARANCE_VALUES,
+  type Appearance,
+  type AppearanceSetting,
+  DEFAULT_SETTINGS,
```
Shreyag02, line 153: A few of these are quite generic for the root namespace of `@raystack/apsara` — `DEFAULT_SETTINGS` especially gives no hint it's theme-related, and `RADII` / `SCALINGS` / `GRAY_PAIRING` are broad too. A `THEME_` prefix or a namespace object would be easier to live with.

Also missing here compared with `theme-preview/index.tsx`: `createThemeScript`, `ThemeScriptParams`, `resolveSettings`, `APPEARANCES`, `ThemeInjectionProps`, `ThemeContextValue`.

packages/raystack/components/avatar/avatar.module.css
```diff
@@ -16,9 +16,8 @@
   --fallback-letter-spacing: 0.03em;
 }
 
-.avatar.avatar-full {
-  border-radius: var(--rs-radius-full);
-}
+/* `radius` comes from the shared override; each `.avatar-size-N` block names
+   its own `--rs-radius-step`. */
```
Shreyag02, line 20: Heads up that all 13 `max(var(--rs-radius-pill, 0px), var(--rs-radius-N))` declarations below are unreachable while `avatar.tsx:128` keeps `defaultVariants.radius: 'medium'` — the (0,2,0) override class always wins. Dropping that default brings them back into play.

packages/raystack/components/theme-preview/__tests__/store.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  it('holds identity across empty results too', () => {
+    expect(readStoredSettings('app')).toBe(readStoredSettings('other'));
+  });
+
+  it('returns the seed as the server snapshot', () => {
+    expect(readServerSettings()).toEqual({});
+  });
+});
+
+describe('writeStoredSettings', () => {
```
Shreyag02, line 90: These cover missing, unparseable and newer-version entries nicely. Could we add one where `localStorage` access itself throws? That's the case behind the comment on `store.ts:111`.

packages/raystack/components/theme-preview/__tests__/script.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      elementId: 'rs-theme-abc'
+    }) as string;
+
+    run(source, element);
+
+    expect(element.getAttribute('data-accent-color')).toBe('orange');
+    expect(element.getAttribute('data-gray-color')).toBe('mauve');
+  });
+
+  it('leaves the server-rendered attribute when the entry is absent', () => {
```
Shreyag02, line 112: This currently encodes the light-first-paint behaviour as expected, so it'll want flipping alongside the `script.ts:69` change — a first visit with OS dark and nothing stored should resolve to `dark`.

packages/raystack/components/theme-preview/store.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/** Reads the settings object out of an entry, tolerating anything at all. */
+function parseEntry(raw: string | null): Record<string, unknown> {
+  if (!raw) return {};
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(raw);
+  } catch {
+    // Includes the legacy `"dark"` bare theme name, which is not an object.
```
Shreyag02, line 55: This comment is the only place the legacy format is acknowledged, and it's handled safely — but the user's saved appearance is silently lost. The old provider stored a bare string (`localStorage.setItem('theme', 'dark')`, default key `theme`), `JSON.parse` throws, and we fall back to defaults.

Either read a bare `light` / `dark` / `system` value once as `appearance`, or note in the migration guide that a fresh `persistKey` is recommended.

packages/raystack/components/drawer/drawer.module.css
```diff
@@ -55,7 +55,8 @@
 .drawerPopup {
   box-sizing: border-box;
   padding: var(--rs-space-3);
-  background-color: var(--rs-color-background-base-primary);
+  background-color: var(--rs-color-panel);
+  backdrop-filter: var(--rs-panel-backdrop-filter);
```
Shreyag02, line 59: Separate from the `-webkit-` point: `Drawer.Content`'s new `radius` prop has no base to work from — this file declares neither `border-radius` nor `--rs-radius-step`.

Measured: `radius="medium"` gives a flat 4px on all four corners of a full-height sheet, including the two flush with the viewport; bare gives 0px. Drawers also don't pick up the theme radius at all.

Per-side radii would solve it, or the prop could be left off Drawer.

packages/raystack/styles/typography.css
```diff
@@ -14,47 +19,49 @@
   /* Font Families */
   --rs-font-title: var(--rs-font-inter), sans-serif;
   --rs-font-body: var(--rs-font-inter), sans-serif;
+  /* JetBrains Mono precedes Menlo: with Menlo first the imported face never
+     rendered on macOS. */
   --rs-font-mono:
-    var(--rs-font-menlo), var(--rs-font-jetbrains-mono), monospace;
+    var(--rs-font-jetbrains-mono), var(--rs-font-menlo), monospace;
```
Shreyag02, line 25: This reorder is the right fix, but it changes all monospace text on macOS for existing `Theme` consumers too. It's covered on the new `ThemePreview` page — worth a line in the release notes as well, since `Theme` users won't have a reason to read that page.

packages/raystack/components/theme-preview/theme-preview.tsx
```diff
@@ -0,0 +1,487 @@
...
+    const previous = previousSettings.current;
+    previousSettings.current = settings;
+    if (!previous) return;
+    const changed: Partial<ThemeSettings> = {};
+    for (const key of THEME_SETTING_KEYS) {
+      if (previous[key] !== settings[key]) {
+        assignSetting(changed, key, settings[key]);
+      }
+    }
+    if (!isSettingsEmpty(changed))
```
Shreyag02, line 347: `onValueChange` fires without a user action on the hydration path.

The `useSyncExternalStore` correction from the server snapshot to the stored value looks identical to a real change here, so consumers get a phantom event on every SSR page load where storage differs from the seed. Anything with a side effect — analytics, a server write, a toast — will fire spuriously.

Verified: CSR gives 0 calls; hydration gives 1, `{"accentColor":"mint"}`.

Gating the first post-hydration reconciliation (e.g. skip while `previousSettings.current` is still the server-derived value) would keep the callback meaning "the user changed something".

packages/raystack/components/theme-preview/__tests__/store.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  });
+
+  it('discards an out-of-union field individually', () => {
+    entries.set(
+      'app',
+      storedEntry({ appearance: 'ultraviolet', radius: 'large' })
+    );
+    expect(readStoredSettings('app')).toEqual({ radius: 'large' });
+  });
+
```
Shreyag02, line 62: The version guard discards a newer-schema entry instead of leaving it alone, so a write from an older build destroys settings a newer build owns.

Verified: with `{v: 99, settings: {appearance: 'dark', radius: 'full'}}` stored, `writeStoredSettings('key', ['accentColor'], {accentColor: 'mint'})` leaves `{v: 1, settings: {accentColor: 'mint'}}`.

Not reachable while `STORAGE_VERSION` is 1, so no rush — but the branch is here to handle exactly this case and currently handles it destructively. Bailing out of the write when the stored `v` is higher would be safer.

Shreyag02 (changes requested):

ravisuhag:
## `reducedMotion` doesn't do anything

`theme.css` collapses the duration tokens:

```css
:where([data-reduced-motion="true"]) { --rs-duration-normal: 1ms; /* … */ }
```

`effects.css` sets them straight back:

```css
:where(:root, [data-theme]) { --rs-duration-normal: 200ms; /* … */ }
```

Both are `:where()`, so both are zero specificity, and `index.css` imports `effects.css` after `theme.css`. A theme element carries both attributes, so `effects.css` wins:

```js
getComputedStyle(el).getPropertyValue('--rs-duration-normal') // "200ms"
```

The `@media (prefers-reduced-motion: reduce)` branch has the same problem — media queries don't add specificity. So the forced path and the system path are both dead, and one of the seven settings is inert.

Moving the two blocks to the bottom of `effects.css` fixes it and keeps the durations in one file.

None of the 1,380 lines of tests catch this, because jsdom doesn't resolve the cascade. A browser check that reads the token off a `data-reduced-motion="true"` element would.

## `onValueChange` — same root cause as your `theme-preview.tsx:347` comment, opposite symptom

Shreyag found it firing when nobody asked. It also never fires for a key the consumer controls.

`setValue` drops controlled keys:

```ts
if (valueRef.current?.[key] !== undefined) continue;
```

and the callback comes from an effect diffing settled settings, which a controlled key can't move. So this does nothing and logs nothing:

```tsx
<ThemePreview value={{ appearance: 'dark' }} onValueChange={console.log}>
  <ThemePreviewSwitcher />
</ThemePreview>
```

Both symptoms come from deriving the callback from state instead of from the request. Firing it inside `setValue` with the requested patch, controlled keys included, fixes both at once — hydration stops producing phantom events because hydration isn't a request, and controlled keys start reporting because they are.

Worth saying the per-key control is a good idea and nothing else has it. Radix Themes has no `onChange` at all, next-themes puts `setTheme` on the hook, Mantine's `forceColorScheme` is an override and is named like one. Being first here is fine. It just means there was no reference implementation to check the contract against.

## 22 components now import from `components/theme-preview/`

```ts
// button.tsx
import { radiusVariants } from '../theme-preview/radius';
// popover.tsx
import { type PortalContainer, useThemeInjection } from '../theme-preview/portal';
```

Button doesn't use the theme. It uses a radius map. `PortalContainer` is a Base UI type restated and has nothing to do with themes either. Both ended up under `theme-preview/` because that's where they got written.

That directory is going to be renamed. Right now that's 22 files we'd have to touch on a day that should only be touching the theme. Moving `radius.ts`, `radius.module.css` and `PortalContainer` somewhere neutral costs a few import lines now.

This sits next to Shreyag's point about `PortalContainer` not being exported. Same type, different problem — it needs a home before it needs an export.

## Smaller

- `styles/primitives/appearance.css` still has `[data-theme="light"], [data-theme="dark"] { transition: background-color 0.4s ease, color 0.4s ease }` unwrapped. Every portal re-injection carries `data-theme` now, so all 13 pick it up. The duration is hardcoded rather than a token, so reduced motion can't touch it even after the fix above. And being unwrapped, a consumer can't override it without `!important`, which is the thing the `:where()` rewrite exists to avoid.

- On the export list, agreed with Shreyag. Two extras: the nine arrays could be the single `SETTING_VALUES` object that already sits in `settings.ts` and isn't exported — a settings picker that maps over it stays correct when an eighth setting arrives. And `radiusClasses` / `radiusVariants` hand out CSS Modules hashed strings. That's build output, and exporting it means we're promising it stays stable. `radiusClass()` is the only one of the three with a real contract.

- `build-no-fonts-css.js` doesn't fail when the regex matches nothing. If the font host changes or postcss stops hoisting the `@import`s, it writes an identical file called `style-no-fonts.css` and consumers load Google Fonts believing they don't. One `if (removed === 0) process.exit(1)` closes it.

- The description calls this additive and it mostly is, but wrapping the token layer in `:where()` changes behaviour for existing `Theme` users too — a consumer rule on `:root` that used to lose to `[data-theme="dark"]` now wins. Same category as the font-mono reorder Shreyag flagged. Probably a release-note line rather than a change.

ravisuhag:
Separate from the bug list, one API question we should settle now rather than after release.

The seven settings arrive as one `value` bag and one `defaultValue` bag. We should flatten the controlled half.

Where the ecosystem sits — the split tracks whether the settings are a closed set of enums or an open tree of tokens:

| Library | Shape | What it passes |
|---|---|---|
| Radix Themes `<Theme>` | Flat | `appearance`, `accentColor`, `grayColor`, `radius`, `scaling`, `panelBackground`, `hasBackground` |
| React Spectrum `<Provider>` | Flat | `theme`, `colorScheme`, `scale`, `locale` |
| next-themes | Flat | `defaultTheme`, `forcedTheme`, `storageKey`, `enableSystem` |
| Mantine | Both | `theme={}` for tokens, flat `defaultColorScheme` / `forceColorScheme` |
| MUI | Object | `theme={createTheme(…)}` |
| Ant Design | Object | `theme={{ token, algorithm, components }}` |

Closed enums go flat. Token trees go in an object. We have seven closed enums, which is the Radix Themes case rather than the MUI one — and we already took `hasBackground` and six of the setting names from it.

Not full flat though, because the two props get used differently.

`defaultValue` is bulk. It's a cookie or a stored object, seven keys at once. A bag fits that.

`value` is one or two keys. A brand accent from app state, a forced dark scope. Flat fits that.

So:

```tsx
<ThemePreview
  defaultValue={cookieSettings}   // bag: bulk seed, one line from storage
  accentColor={brand}             // flat: controlled, usually one key
  onValueChange={handleChange}
  persistKey="app-theme"
/>
```

Mantine ships this exact split, for the same reason.

Two things come along with it. The settings show up in autocomplete and get their own rows in the props table, instead of hiding one level down inside `Partial<ThemeSettings>`. And consumers stop handing us a fresh object literal on every render.

One thing to be clear about before it comes up: this does not let us delete `useStableSettings`. That holds the *computed* settings stable, and that object is rebuilt every render whatever the props look like.

What it costs: seven new prop names, and `defaultValue` sitting next to `accentColor` is a little asymmetric. The docs table, the demo and the playground panel all change. If we'd rather keep the one bag, that's defensible — Radix Themes is flat but has no controlled story at all, so nobody has solved our exact case.

We shouldn't add per-key callbacks. `onAppearanceChange` times seven is seven props for nothing, and `changed` already says which key moved.

The callback fix in the other comment matters more than this one. Without it, flattening just makes a broken contract easier to see.

rohanchkrabrty:
> Separate from the bug list, one API question we should settle now rather than after release.
> 
> The seven settings arrive as one `value` bag and one `defaultValue` bag. We should flatten the controlled half.
> 

@ravisuhag Thought about this approach during the RFC, but discarded it.

Flattening the keys adds noise without much value over the object-based API. Radix UI has 6 separate keys, each taking an object with value and defaultValue, along with 6 callbacks. This makes the API quite noisy. [Source](https://github.com/radix-ui/themes/blob/main/packages/radix-ui-themes/src/components/theme.tsx)

The hybrid approach - flat keys with a single `onChange` callback and a shared `defaultValue`, also feels confusing and goes against our codebase convention of `value`, `defaultValue`, and `onValueChange`.

Let's keep the existing approach. Will address the discoverability by improving the docs and the callback bug

packages/raystack/components/theme/store.ts
```diff
@@ -0,0 +1,138 @@
...
+    const next = patch[key];
+    if (next === undefined || settings[key] === next) continue;
+    settings[key] = next;
+    changed = true;
+  }
+  if (!changed) return;
+
+  const entry: StoredEntry = { v: STORAGE_VERSION, settings };
+  try {
+    window.localStorage.setItem(persistKey, JSON.stringify(entry));
```
rohanchkrabrty: Done. `writeStoredSettings` returns a boolean now, and `setValue` falls back to in-memory state when the write is refused.

packages/raystack/components/theme-preview/theme-preview.tsx
rohanchkrabrty: Done, the fallback lives in `setValue` now.

packages/raystack/components/theme-preview/script.ts
rohanchkrabrty: Done. The script takes the seeded settings and resolves `system` via `matchMedia` when nothing is stored. It's also emitted without a `persistKey` when appearance is `system`, so a first visit on a dark OS paints dark.

packages/raystack/styles/theme.css
rohanchkrabrty: Agreed on the bug. Went with the Radix split rather than a cap: controls keep `max(--rs-radius-pill, …)` (Button, IconButton, Badge, Chip, Input, Avatar, Select trigger, Tabs, Toggle…), every surface drops to its plain `--rs-radius-N`. Radix does the same, `max(var(--radius-N), var(--radius-full))` in base-button, plain `var(--radius-4)` in dialog and card. A cap would give a dialog 16px under `full` but 6px under `large`, a second curve to explain.

packages/raystack/components/avatar/avatar.tsx
rohanchkrabrty: Done, default removed.

apps/www/src/content/docs/theme/preview/index.mdx
rohanchkrabrty: Updated with the script fix, the paragraph now describes the `system` resolution on first visit.

packages/raystack/components/dialog/dialog.module.css
rohanchkrabrty: Done. Added autoprefixer to the rollup postcss chain with a browserslist (`defaults`, Safari 15+). All 14 sites carry `-webkit-backdrop-filter` in the built CSS.

packages/raystack/components/theme-preview/theme-preview.tsx
rohanchkrabrty: Done. It compares the theme's own appearance from any source now: controlled, stored, in-memory or seeded.

packages/raystack/components/theme-preview/portal.ts
rohanchkrabrty: Pulled the `container` prop and `PortalContainer` out of this PR. Unrelated to the theme work, will take it up separately.

packages/raystack/components/theme-preview/index.tsx
rohanchkrabrty: `PortalContainer` is out of this PR (see the portal.ts thread). `createThemeScript` and `ThemeScriptParams` are exported from the package root now.

packages/raystack/index.tsx
rohanchkrabrty: Done. The root exports `THEME_SETTING_VALUES` and `THEME_DEFAULT_SETTINGS` instead of the nine arrays, drops the hashed `radiusClasses` / `radiusVariants` (kept `radiusClass`), and adds `createThemeScript`, `ThemeScriptParams`, `ThemeInjectionProps` and `ThemeContextValue`.

packages/raystack/components/avatar/avatar.module.css
rohanchkrabrty: Done, reachable again with the default gone.

packages/raystack/components/theme/__tests__/store.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  it('holds identity across empty results too', () => {
+    expect(readStoredSettings('app')).toBe(readStoredSettings('other'));
+  });
+
+  it('returns the seed as the server snapshot', () => {
+    expect(readServerSettings()).toEqual({});
+  });
+});
+
+describe('writeStoredSettings', () => {
```
rohanchkrabrty: Added, both for a throwing `setItem` and a throwing `localStorage` accessor.

packages/raystack/components/theme-preview/__tests__/script.test.ts
rohanchkrabrty: Flipped, plus cases for the seed fallback and the no-`persistKey` path.

packages/raystack/components/theme-preview/store.ts
rohanchkrabrty: Going with the docs note. Reading the bare string would need the inline script to accept it too, and RFC 004 already settles this: the bare name fails to parse as an object and resolves to the seed. Migration is a code change anyway, so the consumer picks the `persistKey`. Added a line under "Migrating from Theme".

packages/raystack/components/drawer/drawer.module.css
rohanchkrabrty: Done with per-side radii. The popup has a base step and `--rs-radius-2` now, rounds only the two free corners per `side`, and the flush corners stay square under the `radius` prop as well.

packages/raystack/styles/typography.css
rohanchkrabrty: Reverted the reorder. Menlo first is the intended design (system mono on macOS, JetBrains Mono elsewhere), so there's nothing to note.

packages/raystack/components/theme-preview/theme-preview.tsx
rohanchkrabrty: Fixed from the other end: `onValueChange` fires inside `setValue` from the requested patch, and the settled-state effect is gone. Same root cause as Ravi's controlled-key point. A hydration gate would stop the phantom event but leave controlled keys silent; firing from the request fixes both. Tests cover hydration, a storage-driven change and a controlled key.

packages/raystack/components/theme/__tests__/store.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  });
+
+  it('discards an out-of-union field individually', () => {
+    entries.set(
+      'app',
+      storedEntry({ appearance: 'ultraviolet', radius: 'large' })
+    );
+    expect(readStoredSettings('app')).toEqual({ radius: 'large' });
+  });
+
```
rohanchkrabrty: Done. The write bails with `false` when the stored `v` is newer, and the change falls back to memory.

rohanchkrabrty:
**1. `reducedMotion`** - Moved both blocks to the end of `effects.css`.

**2. `onValueChange`** - Fires from `setValue` with the requested patch. Controlled keys report without applying; hydration and storage changes are silent. Tests for all three.

**3. Imports from `theme-preview/`** - `radius.ts` and its CSS module moved to `shared/radius`, next to `shared/gap`, with `Radius` exported from there. The `container` prop and `PortalContainer` are out of this PR altogether; unrelated to the theme, will take it up separately.

**4. Transition in `appearance.css`** - Duration is `--rs-duration-slow` and the selector is wrapped in `:where()`. Kept it on portals: the colour only moves on an appearance switch, and a popup that snaps while the page fades looks broken. `disableTransitionOnChange` covers both.

**5. Exports** - Dropped `radiusClasses` and `radiusVariants` from the root, kept `radiusClass`. The nine arrays are replaced by `THEME_SETTING_VALUES`, with `THEME_DEFAULT_SETTINGS` beside it.

**6. `build-no-fonts-css.js`** - Exits 1 when the regex matches nothing.

**7. `:where()` and existing `Theme` users** - Agreed, release note. Added a theme entry to the CHANGELOG with this under "Changes for existing Theme users".

@ravisuhag

ravisuhag (approved):

Shreyag02 (approved):

PR #898, opened 2026-09-05
feat!: CalendarPreview scale-aware selection

Shreyag02:
## Summary

stacked on `feat/calendar-preview-rangepicker`.

Adds coarser-than-a-day selection to `CalendarPreview` — month, quarter, half-year and year — behind a tab switcher above the grid.

A `Date` can't say whether it means "August 2026" or "1 August 2026", so above day scale the value becomes a `ScaleValue`:

```ts
{ date: '2026-08-01', scale: 'month' }
```

The scale travels with the value instead of sitting in a separate prop, so a stored value still reads back as August 2026 with no calendar mounted.

It also answers the four API questions #894 deferred to this PR, and closes 27 findings from three rounds of review and browser testing.

> ⚠️ **This PR contains breaking changes.** See [Breaking changes](#breaking-changes) before upgrading.

## Changes

**Feature**

- New root props: `scales`, `trailingValue`, and `scale` / `defaultScale` / `onScaleChange`
- Ten new parts: `.Body`, `.Label`, `.Scales`, `.Scale`, `.Separator`, `.Panel`, plus the period views `.Months`, `.Quarters`, `.HalfYears`, `.Years`
- `trailingValue` makes a period emit its **last** day instead of its first — an end field gets 31 July from "July 2026" where a start field gets the 1st
- `defaultDate` follows the value's shape, and `.Reset` restores the day **and** the scale together
- `.Trigger` takes `nativeButton`, for a `render` prop that produces a real `<button>`

**Fixes most likely to affect you**

- The typed field couldn't be typed into — opening the popover moved focus to the previous-month button, so keystrokes went nowhere. A range now fills entirely from the keyboard
- A scale round trip rewrote the field: day → year → day turned a committed 15 August into "01 Jan 2026", with nothing marked in the grid
- A period cell lit up for the wrong scale — typing `H1 2028` marked **Q1 2028**
- A date typed into an empty **End** field landed in **Start**
- `.Reset` dropped focus to `<body>`, and a `.Trigger` without an `.Input` had no tab stop at all — neither was reachable by keyboard
- Period lists no longer render years with nothing selectable in them (half the cells under a mid-range `minDate`)
- A rejected endpoint stayed marked invalid after its partner moved and made it valid again

**Docs** — `props.ts` and `index.mdx` brought back in step with the component, plus demo tabs for scale selection and `trailingValue`.

### Breaking changes

| Change | What to do |
|---|---|
| `.Reset` marks "nothing left to restore" with `aria-disabled="true"` instead of the `disabled` attribute | Anything keyed on `:disabled`, `[disabled]` or `toBeDisabled()` for that state wants `[aria-disabled]`. A `disabled` / `readOnly` root, and a caller's own `disabled` prop, still render a real `disabled` |
| `CalendarPreview.Picker` is renamed `.Body` | Rename the part |
| `data-scale` is no longer stamped on `.Body`, `.Caption`, `.Days`, `.Input`, `.Label`, `.Trigger` or the period view containers | It stays on the root, on day and period cells, on `.Panel` and on `.Scale` — retarget selectors there |
| `scales` keeps the order it's given rather than being sorted | `scales={['month','day']}` now opens on month; `scales[0]` is the default |
| `onScaleChange` fires when a draft is dropped (Escape, `.Reset`) | The scale really did change, so it's reported — but you'll see a call you didn't before |
| `useCalendar().value` widens to include `ScaleValue` | Narrow before using it as a `Date` |

Smaller, DOM-visible but unlikely to bite: a period cell carries `data-selected` only at its own scale, and `.Days` returns `null` when the scale isn't `'day'` so it can sit beside the period views in `.Panel`.

### Why three prop arms

TypeScript can't inspect an array's contents, so the arms discriminate on the **shape** of `scales`: omitted or the literal `'day'` keeps `Date`; any other scale, or any array, moves to `ScaleValue`. Known wart — `scales={['day']}` takes the scale-aware arm where `scales='day'` doesn't. It's documented on the type rather than worked around.

### Bounds at period scales

`isAvailable` tests the day a period would **produce**, not the period itself, so the same period answers differently at each end of a pair. Bounded at 15 July 2026: Q3 2026 is disabled for a start field (emits 1 July) and available for an end field (emits 30 September).

### `isDateUnavailable` is day scale only

A day predicate has no single lift to a period — one blocked day blocking all of August is as wrong as it not blocking it — and calling it per cell would run it 365 times a year. Period cells are bounded by `minDate` / `maxDate` instead. Now stated in the prop's JSDoc.

### Scale drafts

Switching scale never emits; it sets a draft that a cell click or Enter commits and Escape drops. A run of switches converts from where the run **started**, not from the draft it last produced — `convertScale` is lossy outward and doesn't undo, which is what turned 15 August into 1 January.

### `ChangeDetails` is one shape

Every arm reports the same `{ reason, period, toDate }`, rather than a per-arm union, so a consumer never branches on the arm to read the details. `toDate()` is never null — a clear still says which day it cleared. This is a deviation from RFC 005, which proposed case-specific payloads; it isn't recorded there yet.

## Test Plan

- [x] Manual testing completed — every arm driven end to end in a **real browser**, not jsdom: single-day select / clear / bounds / `isDateUnavailable` / `readOnly` / `disabled`; the range machine including restart, same-day, out-of-order and read-only endpoints; scale switching, both period edges, Escape, all seven typed input forms; focus and dismissal on all three trigger shapes; light and dark; 420px viewport. Every defect was reproduced in the browser before being fixed and re-verified after
- [x] Build and type checking passes — `pnpm --filter @raystack/apsara build`, `tsc --noEmit` clean for `calendar-preview`, biome clean on every file touched
- [x] **604 tests** in the calendar-preview family (88 in `scale-selection.test.tsx`), **3311** across the package
- [x] Demo page loads with zero console errors

## SQL Safety

Not applicable — this repo is TypeScript and CSS; the PR touches no Go or SQL.


rohanchkrabrty (changes requested):
1. **Typed day on a scale-aware root emits a `Date`.** `calendar-preview-input.tsx:206` calls `setValue` with a raw `Date`, while a click wraps it as `{ date, scale: 'day' }` at `calendar-preview-root.tsx:463`. Route the typed path through the root so the input never builds the value shape.

2. **Focus-to-open stops working after Escape or a trigger press.** `root.tsx:390-410` arms a flag that is only cleared by the next focus event. In `<Trigger><Input/></Trigger>` focus is already in the input, so no focus event comes and the flag stays armed. `trigger.tsx:77` returns on `pressing` before consuming it, and `pressing` never resets if the pointer is released outside. Arm only when focus is outside the trigger at close time.

3. **An uncommitted scale draft shows in the closed field.** Nothing clears `scaleDraft` on close. Switch to Quarter, click outside: `value` is unchanged but the input reads `Q3 2026` and the root has `data-scale="quarter"`. Escape is handled only in `body.tsx:35`, so `<Content><Panel/></Content>` never drops a draft. Handle commit or drop in the root's `setOpen` by close reason.

4. **Escape and Reset bypass `onScaleChange`.** `root.tsx:550` and `root.tsx:590` use `setScaleUnwrapped`. Controlled `scale` gets stuck on the drafted scale. Use `setScale`.

5. **Input and Trigger format a period differently.** `input.tsx:216` uses the view scale, `trigger.tsx:101` uses the value's scale. A quarter value with `defaultScale='day'` shows `Q3 2026` in the trigger and `01 Jul 2026` in the input. Also `defaultFormatValue` never receives `timeZone`.

6. **The trigger div is `role="button" tabindex="0"` around an `<input>`.** Base UI adds both for non-native triggers. Result: two tab stops per field, four for a range, and interactive content inside a button role. Set `tabIndex: -1` and no role when the trigger wraps an input, or make the input itself the trigger.

7. **Input Validity is read from a ref during render** at `input.tsx:248`. This should be state driven since ref can go stale

8. **`parseScaleInput` ignores the injected `today`** at `input.tsx:156`, so a bare `Q4` uses the wall clock. The parser's `trailing` option is dead code.

9. **Typed commits close inconsistently.** A typed `Q4 2026` closes the popover, a typed day leaves it open.

10. **Clearing one range input clears the whole range** at `input.tsx:194` and leaves the old draft painted in the grid.

11. **`'scale'` as a change reason is never emitted.**

12. **`useCalendar()` does not expose the draft**, though the RFC says it does. `CalendarPreviewDraftRange` is exported with nothing returning it.

13. **`.Scale` is a plain button, not a tab.** `scales.tsx:77-107` has no `aria-selected` and no roving focus, unlike the default `Tabs`. `root.tsx:325` also canonicalises order, so reorder via `scales` does not work.

14. **Slim the date adapter.** Keep the `DayKey` core and `zoned()`. Delete `epoch`, `formatCaptionLabel`, `monthStart` and the six `startOf*/endOf*Key` wrappers, and call date-fns directly. date-fns is plugin-free and already shipped by react-day-picker, so the wrapper does not deliver the isolation or bundle savings the RFC claims. Nothing enforces the single-import rule.

15. **Simplify `lib/scale.ts`.** Accept `DayKey` only, use an options object for `isAvailable`, and use one approach for all scales instead of string maths for half-year and date-fns for the rest.

16. **Drop the context generic and the Provider wrappers.** The value type is a closed union now; twelve parts repeat the same cast.

17. **Remove stale comments and dead code.** Comments at `root.tsx:319`, `grid.tsx:97`, `grid.tsx:382`, `picker.test.tsx:166`, `calendar-preview.test.tsx:1017` describe shipped phases. `.range-fields` CSS is unused. `root.tsx:537` casts a valid value `as never`. `data-scale` is stamped on 14 elements where root plus cells is enough.

18. **Period lists are expensive.** The month view renders 252 buttons and each does about six date-fns calls per render, unmemoised, on every context change. Memoise the cell list and compute the bound keys once.

19. **A Tooltip root per day cell**, 84 in a two-month range picker, even with tooltips off. Mount it only when `showTooltip` is on.

20. **`vitest.config.mjs` pins `TZ=UTC` for the whole package** to fix one test. It hides zone bugs everywhere else. Fix that test instead.


Also we need to create a separate issue and remove `dayjs` across the library

apps/www/src/content/docs/components/calendar-preview/index.mdx
rohanchkrabrty: Remove this, and similar comments from the docs

rohanchkrabrty (changes requested):
1. the chevron button sizes/padding/icon sizes and the month selector button doesn't match design 

Image

2. the scrollbar should be at the outer edge to prevent overlap with content.

Image

3. with showWeekNumber and fixedWeeks enabled, there is a trailing week number that shouldn't exist

Image

4. popup keeps on opening even on outside click. if input is invalid it doesn't change even after i select a new date

https://github.com/user-attachments/assets/d86cc233-4359-4aa7-9b13-2675de0d2273

5. The calendar instantly closes on range selection, which shouldn't happen.

6. With disabled dates, a range shouldn't be allowed to select over them

Image

7. In range readOnly start example I couldn't change the end date

8. If I open the range calendar popup with start trigger and click on end trigger , the pop gets closed. Clicking in between the triggers shouldnt close the popup

9. The scale selection tabs doesn't match design

10. the grid cells except for scale-day, doesnt match design

11. for the scale year, there should be only cells and no header for year

Image

Image

Shreyag02:
Thanks both — working through all of it. Six of the findings describe code that has since changed, so I want to flag those rather than have you re-read them.

**Already addressed on the current head:**

1. **Typed day on a scale-aware root emits a `Date`.** Days no longer become values in the input. `commitDay` on the root is the single place a day is written, and it wraps as `{ date, scale: 'day' }` whenever the root is scale-aware. The input just calls it.

2. **Escape and Reset bypass `onScaleChange`.** Both route through `settleScale` → `setScale`, so `onScaleChange` fires when a draft is dropped, the same as when one is set. The raw setter appears only at its declaration and inside `setScale`.

3. **Input and Trigger format a period differently.** The input reads the endpoint's own scale now — `formatValue(endpoint, isScaleValue(endpoint) ? endpoint.scale : scale)` — so a committed quarter reads back as a quarter while the view sits on days. Matches what `.Trigger` does. The other half of that finding stands, though: `defaultFormatValue` still takes only `(value, scale)` and never receives `timeZone`. I'm counting that one as open.

4. **A valid value cast `as never`.** Gone; there are no `as never` casts left in non-test source.

5. **`props.ts` value types are missing the period shape.** `{ date: string; scale: Scale }` is in `value`, `defaultValue`, `onValueChange` and `formatValue`.

6. **`index.mdx` says the scale setter arrives in a later phase.** It now reads that the scale is read-only by decision — switching is `.Scales` and `.Scale`.

@rohanchkrabrty — items 1 to 4 were already in at the commit your review is pinned to, so I assume the reading started before I pushed. My fault for not noting the push on the PR at the time.

Everything else in both reviews I've confirmed against the current head and am working through. I'll reply per-thread as each lands, and I'll flag separately the few that I think are product decisions rather than defects (the auto-close on range completion, and whether emptying one endpoint should clear the whole range).

Shreyag02:
**On "with disabled dates, a range shouldn't be allowed to select over them"** — fixed, and here is the reasoning behind the shape of it, since there was more than one defensible answer.

**What the behaviour is now.** Both endpoints were already checked; the days between them were not, so a range could bracket a day marked unavailable and hand it back inside the value. Two paths, two answers:

- A **click** that would complete over a blocked day restarts from that day instead of completing. That is the arm the from/to machine already has for a click it cannot use — an earlier second click behaves the same way — so it introduces no new concept.
- A **typed** endpoint cannot restart, because typing names the field it lands in. It is rejected with `unavailable`, which is the reason `.Input` already reports for a typed date that lands on a blocked day. The field shows its error as it does for any other rejection.

**Bounds are deliberately not part of this.** `minDate` and `maxDate` describe a single contiguous window, so two endpoints inside it cannot straddle anything outside it — bounds are incapable of leaving a hole in the middle. Only `isDateUnavailable` can, so that is the only predicate the span check consults.

**No prop guards it.** react-day-picker makes the equivalent opt-in via `excludeDisabled`, and the existing `RangePicker` does not pass it, so today a range there may span disabled days. I did not copy that default: `excludeDisabled` arrived in a patch release and reads as backwards compatibility, and the old family never exposed per-date disabling at all, so there is no in-repo behaviour being preserved. A day the consumer has declared unselectable, returned inside a range, is hard to defend as correct. If someone later needs "not selectable as an endpoint, but spannable", an opt-out is additive and breaks nobody.

**One judgement worth naming**, in case you read it differently: I have taken "a blocked day inside an emitted range" as selecting it. That reading is what makes prevention the default rather than a prop.

Covered by three tests — a click restarting rather than completing, a range that clears the blocked day still completing, and a typed endpoint rejected with `unavailable`. Each fails with its guard removed.

apps/www/src/content/docs/components/calendar-preview/index.mdx
Shreyag02: Done — the preview / semver callout at the top is gone, along with the phase hedging further down. The one callout left is about a failed commit keeping the typed text, which is behaviour rather than a disclaimer, but happy to cut that too if you'd rather.

packages/raystack/components/calendar-preview/calendar-preview-trigger.tsx
```diff
@@ -54,28 +65,60 @@ export function CalendarPreviewTrigger({
...
         'data-scale': scale,
+        /* Base UI gives a non-native trigger both, which around a field is a
+           second tab stop and a control inside a button role. */
+        role: hasInput ? undefined : 'button',
+        tabIndex: hasInput ? -1 : undefined,
+        /* Merged to the right of `useClick`, so this runs first. Only the
+           closing half goes, or a press could not reopen a field that never
+           lost focus. */
+        onClick: (event: BaseUIEvent<MouseEvent<HTMLDivElement>>) => {
+          if (hasInput && open) event.preventBaseUIHandler();
```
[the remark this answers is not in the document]
Shreyag02: This is already scoped to the trigger. Inputs register through a context provided inside the trigger's own children, so an `.Input` mounted in `.Content > .Body` never registers with a childless trigger sitting beside it — that trigger keeps its button role, and its click still closes the popover.

rohanchkrabrty (changes requested):
Unresolved comments from before
- **12 - useCalendar() does not expose the draft**

- **1 - month selctor button doesnt match design** - The grey background is a hover/focus state, not a default state

Image

- **2 - the scrollbar should be at the outer edge to prevent overlap with content.** - This is not fixed and instead the scrollbar is not visible anymore
- **4 - popup keeps on opening even on outside click.**  - The popup is fixed, but the input regains focus on clicking outside for the same scenario. After opening and clicking outside, the input should not gain focus
- **9 - The scale selection tabs doesn't match design**
- **10 - the grid cells except for scale-day, doesnt match design**


New bug 
- `onScaleChange` fires twice on Escape

rohanchkrabrty (approved):

PR #899, opened 2026-09-06
docs: restructure page hierarchy in docs

ravisuhag:
Docs headings were inconsistent, and on the bigger pages they got in the way. DataView had 40 headings on one page, with Timeline's prop tables 250 lines away from the prose explaining them. Several pages had catch-all sections — `Notes`, `Important Notes`, `Features` — that collected whatever didn't fit elsewhere. And a lot of descriptions just restated their own heading:

> **Color** — The Separator component supports three colors.
> **Size** — The Separator component supports three sizes.

This settles on one section order and applies it across all 87 pages.

## Section order

`Overview` (only when the name doesn't explain it) → `Anatomy` → `Usage` → `API Reference` → `Accessibility`

Three rules underneath it:

- Two heading levels. `h4` only inside API Reference, for a sub-type directly under its parent part.
- A page splits only when a part has its own concepts, its own props and its own examples.
- One `h3` answers one question. Appearance galleries before behaviour recipes.

Usage moves above API Reference. The hero demo is static and the playground sits behind a modal, so nothing above the prop tables shows you how to use the component — you hit a wall of types before a single working example. Anatomy and API Reference also overlap 64% by name, so the old order covered the same ground twice before showing anything new.

## What changed

- **DataView split** into `index` / `list` / `timeline`. Timeline's props now sit with the prose that explains them. Server mode folds into `index` — it has no props of its own, so it doesn't meet the bar for its own page.
- **69 pages reordered.** `Examples` is gone as a heading.
- **Six catch-all `h2`s** moved to where they belong. `form`'s Usage went from 11 jumbled entries to an order that follows how you'd actually build a form.
- **~145 headings** normalised to sentence case, keeping component names, acronyms and code spans intact.
- **~80 descriptions rewritten**, with every value read from `props.ts`.

## Fixes

- Button's `colorsDemo` rendered every "Accent" swatch as `color="success"` — Accent and Success looked identical in all four variant tabs.
- Dialog's "With body and footer" was described as *"Example with header and body."*
- Navbar had `Accessibility` as both an `h3` and an `h2`, which made `#accessibility` ambiguous.
- Two broken anchor links.

## Also in here

Button gets a **"render as another element"** example. The `render` prop shipped with no demo.

`sidebar.tsx` renders a split component as a group that opens only while you're inside it, with the index row labelled "Overview" instead of repeating the component name. It's the only non-content file in the diff.

## Checks

`next build` passes, 187 static pages, all 87 docs routes prerendered. Every internal anchor link resolves. `<Demo>` and `auto-type-table` counts were diffed against `main` on every page to confirm nothing was dropped.

### DIFF
--- a/apps/www/src/components/docs/sidebar.tsx
+++ b/apps/www/src/components/docs/sidebar.tsx
@@ -25,19 +25,30 @@ function renderNode(node: Node, pathname: string): ReactNode {
     node.children &&
     node.children.length > 0
   ) {
+    // One component documented across several pages. Two things keep it from
+    // disturbing the flat component list it sits in: the index renders as
+    // "Overview" (the group label already names the component, so repeating it
+    // reads as a duplicate row), and the group only opens while the reader is
+    // inside it. `isInside` is folded into the key so entering or leaving the
+    // section remounts the group at the right default, while leaving the
+    // reader free to toggle it by hand once they're there.
+    const isInside = isActiveUrl(node.index.url, pathname);
     return (
       <Sidebar.Group
         label={node.name as string}
-        key={node?.$id}
+        key={`${node?.$id}-${isInside}`}
         leadingIcon={node?.icon}
+        collapsible
+        defaultOpen={isInside}
         classNames={{
           items: styles.items,
           label: styles.label
         }}
       >
-        {/* Render index item first */}
-        {renderNode(node.index, pathname)}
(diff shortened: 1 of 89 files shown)


ravisuhag:
Stacked on top of this: **#900** — docs typeset + hardcoded type cleanup. Review this one first; #900's diff is scoped to the typography work.

